PR #9, opened 2022-11-22
feat: added signature subscribe

Lioncat2002:


src/rpc.rs
```diff
@@ -114,6 +114,23 @@ impl LightRpcRequestProcessor {
         )
     }
 
+    fn subscribe_signature(
+        websocket_url: &String,
+        signature: &Signature,
+        commitment:CommitmentLevel
+    ) -> std::result::Result<SignatureSubscription, PubsubClientError> {
+        PubsubClient::signature_subscribe(
+            websocket_url.as_str(),
```
godmodegalactus, line 123: Then  you should be able to treat the message you receive from websockets and update the signature status in the map.

godmodegalactus (changes requested):

src/rpc.rs
godmodegalactus: let me know if you have created another issue for this.

Lioncat2002: yeah I will create a new issue for that

godmodegalactus (approved):

PR #14, opened 2022-11-30
Adding ws subscriptions

godmodegalactus:
Implementing websockets for onSignature and onSlot

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -30,7 +30,7 @@ pub fn build_args<'a, 'b>(version: &'b str) -> App<'a, 'b> {
         .about("a lite version of solana rpc to send and confirm transactions")
         .version(version)
         .arg(
-            Arg::with_name("json_rpc_url")
+            Arg::with_name("json-rpc-url")
                 .short("u")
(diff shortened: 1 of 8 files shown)


src/cli.rs
```diff
@@ -30,7 +30,7 @@ pub fn build_args<'a, 'b>(version: &'b str) -> App<'a, 'b> {
         .about("a lite version of solana rpc to send and confirm transactions")
         .version(version)
         .arg(
-            Arg::with_name("json_rpc_url")
+            Arg::with_name("json-rpc-url")
                 .short("u")
                 .long("url")
                 .value_name("URL_OR_MONIKER")
```
Lioncat2002, line 36: Perhaps using the derive feature of clap would be better for the cli.
But ig Aniket already suggested that?

Lioncat2002: Other than that everything looks good imo

godmodegalactus: I tried to implement by comments of aniket, I have issue to correctly find the default port of websocket.

PR #20, opened 2022-12-10
Fixed typescript tests and revert the cli

Lioncat2002:


Lioncat2002:
The `solana-validator` doesn't have any kill process so the ts test cannot exit after a successful passing
that causes it to hang in a infinite loop
This problem can be avaoid by running the solana-validator on a seperate terminal manually first and then running the ts tests
image

Lioncat2002:
The scripts folder is temporary and needed to run the `solana-validator` via yarn

godmodegalactus:
Looks good. Do we really need all these scripts? Why did you commit these scripts?

godmodegalactus (approved):

PR #22, opened 2022-12-10
sending transactions in batches in a new thread

godmodegalactus:


aniketfuryrocks:
I think it should be more of retry transaction for the max_retries in the send transaction config

aniketfuryrocks:
@godmodegalactus I will take this over

PR #40, opened 2023-01-21
WIP: Prometheus

aniketfuryrocks:


prometheus.yml
```diff
@@ -0,0 +1,13 @@
+global:
```
riordanp, line 1: This file is not required for the deployment, prometheus is already configured

src/workers/prometheus.rs
```diff
@@ -0,0 +1,59 @@
...
+        format!(
+            "HTTP/1.1 200 OK\r\nContent-Length: {}\r\n\r\n{}",
+            payload.len(),
+            payload
+        )
+    }
+
+    async fn handle_stream(&self, stream: &mut TcpStream) -> anyhow::Result<()> {
+        let metrics = self.metrics_capture.get_metrics().await;
+        let metrics = serde_prometheus::to_string(&metrics, None, HashMap::new())?;
```
riordanp, line 31: Please add "literpc" as a prefix so we can namespace metrics https://github.com/w4/serde_prometheus/blob/master/src/lib.rs#L89

src/workers/prometheus.rs
```diff
@@ -0,0 +1,59 @@
...
+
+        stream.flush().await?;
+
+        Ok(())
+    }
+
+    pub fn sync(self) -> JoinHandle<anyhow::Result<()>> {
+        #[allow(unreachable_code)]
+        tokio::spawn(async move {
+            let listener = TcpListener::bind("[::]:9500").await?;
```
riordanp, line 46: You can uncomment the metrics block in fly.toml and change the ports to match which will enable prometheus polling

riordanp (changes requested):
Unfortunately serde_prometheus doesn't seem to support metric types. We'll have to see how these appear in grafana and might need a different solution if it's not working.

prometheus.yml
aniketfuryrocks: Removed

src/workers/prometheus.rs
aniketfuryrocks: Done

src/workers/prometheus.rs
aniketfuryrocks: Fixed

PR #41, opened 2023-01-29
postgres and prometheus and unbounded tx channels

aniketfuryrocks:
Implements prometheus and postgres

Includes
#38 #40


aniketfuryrocks:
env variables needed to implemented

src/workers/tx_sender.rs
```diff
@@ -42,53 +50,54 @@ impl Default for TxProps {
...
+                    warn!("{err}");
+                    0
+                }
+            };
+
+            if let Some(postgres) = postgres {
+                for (sig, recent_slot) in sigs_and_slots {
+                    postgres
+                        .send(PostgresMsg::PostgresTx(PostgresTx {
+                            signature: sig.clone(),
```
godmodegalactus, line 92: This should not be done here, but in block subscription stage where we iterate through all the transactions to find if they are in the block or not.

godmodegalactus (changes requested):

godmodegalactus (approved):

PR #42, opened 2023-01-31
Postgres env variables

aniketfuryrocks:


src/cli.rs
```diff
@@ -29,5 +29,5 @@ pub struct Args {
     pub clean_interval_ms: u64,
     /// addr to postgres
     #[arg(short = 'p', long)]
-    pub postgres_config: Option<String>,
+    pub enable_postgres: bool
```
riordanp, line 32: Dockerfile should be updated with this option for the run command, or it should move to env like the RPC URLs

riordanp (changes requested):

src/cli.rs
aniketfuryrocks: Fixed

riordanp (approved):

PR #50, opened 2023-02-07
Poll Blocks instead of websockets

aniketfuryrocks:
- add MAX_RECENT_BLOCKHASHES (fix block height)
- poll `getBlocks`
- use memo in tests for tx
- reduce cleaner log
- fix: signature subscribe with commitment
- partial rpc send and confirm test
- Refactored BenchHelper


src/workers/block_listenser.rs
```diff
@@ -284,15 +284,44 @@ impl BlockListener {
...
+                    .block_store
+                    .get_latest_block_info(commitment_config)
+                    .await;
+
+                let block_slots = self
+                    .rpc_client
+                    .get_blocks_with_commitment(latest_slot, None, commitment_config)
+                    .await?;
+
+                let block_future_handlers = block_slots.into_iter().map(|slot| {
```
mschneider, line 308: this will spawn too many requests in parallel and overload the rpc. potentially causing missing slots in your block_store

src/workers/block_listenser.rs
```diff
@@ -284,15 +284,44 @@ impl BlockListener {
...
+                            .await
+                        {
+                            warn!(
+                                "Error while indexing {commitment_config:?} block with slot {slot} {err}"
+                            );
+                        };
+                    })
+                });
+
+                let _ = try_join_all(block_future_handlers).await;
```
mschneider, line 324: no error handling at this point will cause missing slots in your block_store

tests/part_send_n_confirm.rs
```diff
@@ -0,0 +1,31 @@
+use bench::helpers::BenchHelper;
+use lite_rpc::{DEFAULT_LITE_RPC_ADDR, DEFAULT_RPC_ADDR};
+use solana_rpc_client::nonblocking::rpc_client::RpcClient;
+
+#[tokio::test]
+async fn part_send_n_confirm() -> anyhow::Result<()> {
+    let rpc_client = RpcClient::new(DEFAULT_RPC_ADDR.to_string());
+    let lite_rpc_client = RpcClient::new(DEFAULT_LITE_RPC_ADDR.to_string());
+
+    send_and_confirm_memo(&rpc_client, &lite_rpc_client).await?;
```
mschneider, line 10: if a tx is not sent via lit_rpc it can not be confirmed via lite_rpc

tests/client.test.ts
```diff
@@ -1,27 +1,31 @@
...
     transaction.add(
-        SystemProgram.transfer({
-            fromPubkey: payer.publicKey,
-            toPubkey: toAccount,
-            lamports: LAMPORTS_PER_SOL,
-        }),
+        new TransactionInstruction({
+            programId: MEMO_PROGRAM_ID,
+            keys: [],
+            data: Buffer.from("Hello")
```
mschneider, line 22: buffer needs to contain random data so you don't end up creating the same tx twice

tests/part_send_n_confirm.rs
```diff
@@ -4,10 +4,14 @@ use solana_rpc_client::nonblocking::rpc_client::RpcClient;
...
 #[tokio::test]
 async fn part_send_n_confirm() -> anyhow::Result<()> {
+    tracing_subscriber::fmt::init();
+    log::info!("rpc {DEFAULT_RPC_ADDR} lite {DEFAULT_LITE_RPC_ADDR}");
+
     let rpc_client = RpcClient::new(DEFAULT_RPC_ADDR.to_string());
     let lite_rpc_client = RpcClient::new(DEFAULT_LITE_RPC_ADDR.to_string());
 
-    send_and_confirm_memo(&rpc_client, &lite_rpc_client).await?;
+    send_and_confirm_memo(&lite_rpc_client, &lite_rpc_client).await?;
```
mschneider, line 13: memo buffer needs to contain random data so you don't end up creating the same tx twice

aniketfuryrocks: yeah internally I am using `uuid` v4 for random string

godmodegalactus:
This PR is outdated. @aniketfuryrocks Could you create a new one with the changes you want to merge.

PR #54, opened 2023-02-13
Setting runtime at start

godmodegalactus:
runtime workers set to 8.


src/main.rs
```diff
@@ -53,6 +54,14 @@ pub async fn main() -> anyhow::Result<()> {
 
     let identity = get_identity_keypair(&identity_keypair).await;
 
+    let runtime = Builder::new_multi_thread()
+        .worker_threads(8)
```
mschneider, line 58: probably should remove #[tokio::main] else you end up with two runtimes

PR #55, opened 2023-02-13
Poll blocks and devide tasks to multiple tasks using channels

aniketfuryrocks:


src/workers/tx_sender.rs
```diff
@@ -60,7 +58,57 @@ impl TxSender {
...
+                0
+            }
+        };
+
+        if let Some(postgres) = postgres {
+            let forwarded_slot = tpu_client.get_tpu_client().await.estimated_current_slot();
+
+            for (sig, recent_slot) in sigs_and_slots {
+                postgres
+                    .send(PostgresMsg::PostgresTx(PostgresTx {
```
mschneider, line 101: no matter how much you try to re-batch on the other end of this channel, queueing all the individual inserts as their own message is the issue.
please follow the tx-backup example and generate batch insert queries: 1 query, inserting as many data tuples as you have.

you can then proceed to revert all this batching logic from the client-side, it's unnecessarily complex. the guiding principle is: when you receive the data it is already batched:

```
        sigs_and_slots: Vec<(String, u64)>,
        txs: Vec<WireTransaction>,
```

mschneider: send_tx should be send_tsx and receive a Vec<PostgresTx>
update_tx should be update_tsx and receive a Vec<PostgresUpdateTx>, which should contain signatures.

src/lib.rs
```diff
@@ -20,7 +20,7 @@ pub const DEFAULT_WS_ADDR: &str = "ws://0.0.0.0:8900";
 #[from_env]
 pub const DEFAULT_TX_MAX_RETRIES: u16 = 1;
 #[from_env]
-pub const DEFAULT_TX_BATCH_SIZE: usize = 128;
+pub const DEFAULT_TX_BATCH_SIZE: usize = 1 << 5;
```
godmodegalactus, line 23: why are you stuck on 1<<5 , 128 value is more lisible and 1<<5 is 32 not 128.

aniketfuryrocks: Yeah as we discussed we don't want a higher batch size.

src/block_store.rs
```diff
@@ -65,48 +65,58 @@ impl BlockStore {
...
-        if last_recent_block.slot < block_info.slot {
-            *self
-                .get_latest_blockinfo_lock(commitment_config)
-                .write()
-                .await = block_info;
+        let slot = block_info.slot;
+        self.blocks.insert(blockhash.clone(), block_info);
+
+        let latest_block = self.get_latest_block_arc(commitment_config);
+        if slot > latest_block.read().await.1.slot {
```
godmodegalactus, line 118: I prefer explicitly having seperate context for .read() and .write()

src/main.rs
```diff
@@ -84,3 +61,22 @@ pub async fn main() -> anyhow::Result<()> {
...
+
+async fn get_identity_keypair(identity_path: Option<String>) -> anyhow::Result<Keypair> {
+    let identity_bytes = if let Some(identity_path) = identity_path {
+        let identity_file = tokio::fs::read_to_string(identity_path)
+            .await
+            .context("Cannot find the identity file provided")?;
+
+        identity_file
+    } else {
+        let Ok(identity_from_env) = env::var("TPU_IDENTITY") else {
```
godmodegalactus, line 73: Should be "IDENTITY" why break the dependency.

src/main.rs
```diff
@@ -84,3 +61,22 @@ pub async fn main() -> anyhow::Result<()> {
...
+            return Ok(Keypair::new());
+        };
+
+        identity_from_env
```
godmodegalactus, line 77: IDENTITY variable could be pointing to a file

src/tpu_manager.rs
```diff
@@ -27,7 +27,7 @@ pub struct TpuManager {
     error_count: Arc<AtomicU32>,
     rpc_client: Arc<RpcClient>,
     // why arc twice / one is so that we clone rwlock and other so that we can clone tpu client
-    tpu_client: Arc<RwLock<Arc<QuicTpuClient>>>,
```
godmodegalactus, line 30: I think you dont understand what i was doing here. First thing is that you are exposing RwLockReadGuard<QuicTpuClient> as pub which is a big mistake. I have explicitly added comment why there are two arc. With your changes when the connection is being reseted it has to wait for all the other read locks to finish reading to reset the connection. That is not we want, when we reset the connection it will directly update the TPUClient.

Please dont do things that you dont understand.

src/tpu_manager.rs
```diff
@@ -94,15 +94,15 @@ impl TpuManager {
...
+            *self.tpu_client.write().await = tpu_client;
             info!("TPU Reset after 5 errors");
         }
 
         Ok(())
     }
 
-    async fn get_tpu_client(&self) -> Arc<QuicTpuClient> {
-        self.tpu_client.read().await.clone()
+    pub async fn get_tpu_client(&self) -> RwLockReadGuard<QuicTpuClient> {
```
godmodegalactus, line 104: Should never ever expose guards outside of the class. I think I had explicitly told you so.

godmodegalactus (changes requested):

Cargo.toml
```diff
@@ -43,4 +43,5 @@ native-tls = "0.2.11"
 postgres-native-tls = "0.5.0"
 prometheus = "0.13.3"
 lazy_static = "1.4.0"
-dotenv = "0.15.0"
\ No newline at end of file
+flume = "0.10.14"
```
mschneider, line 46: why introduce another dependency to our stack? why not use crossbeam / async channel?

aniketfuryrocks: this is a mpmc channel. We need a simple spmc solution to distribute tasks across multiple threads.
Crossbeam doesn't work great with tokio as discussed with galactus. Therefore `flume`

src/lib.rs
godmodegalactus: Revert this back

godmodegalactus:
@aniketfuryrocks 
This PR is too big, could you just create smaller PR's with just intended changes.

Cargo.toml
aniketfuryrocks: Can replace this with async_channel

PR #62, opened 2023-02-22
use join_all

aniketfuryrocks:


src/workers/block_listenser.rs
```diff
@@ -296,94 +294,60 @@ impl BlockListener {
...
+                } else {
+                    slot_que.append(&mut new_block_slots);
+                    slot_que.split_off(slot_que.len().min(16))
+                };
+
+                let index_futs = slots_to_get_blocks
+                    .into_iter()
+                    .map(|slot| self.index_slot(slot, commitment_config, postgres.clone()));
+
+                join_all(index_futs).await;
```
godmodegalactus, line 350: Before we retried if the fetch of block failed for some reason. You should implement a retry logic.

src/workers/block_listenser.rs
```diff
@@ -296,94 +294,60 @@ impl BlockListener {
...
-                .slot;
-            // -5 for warmup
-            let mut last_latest_slot = last_latest_slot - 5;
-
-            // storage for recent slots processed
-            let rpc_client = rpc_client.clone();
             loop {
-                let new_slot = rpc_client
-                    .get_slot_with_commitment(commitment_config)
+                let mut new_block_slots = self
```
godmodegalactus, line 314: You can remove get blocks and move to the previous implementation of get slot and then create a range of last_slot to current slot. 
Some time i feel getBlocks does not work very well.

src/workers/block_listenser.rs
```diff
@@ -296,94 +294,60 @@ impl BlockListener {
...
-                {
-                    let mut lock = slots_task_queue.lock().await;
-                    for slot in new_block_slots {
-                        lock.push_back((slot, 0));
-                    }
+                let new_latest_slot = *new_block_slots.last().unwrap();
+
+                if latest_slot == new_latest_slot {
+                    warn!("No new slots for {latest_slot}");
+                    tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
```
godmodegalactus, line 329: Reduce this wait to 200ms.

src/workers/block_listenser.rs
```diff
@@ -296,94 +294,60 @@ impl BlockListener {
...
-                let new_block_slots: Vec<u64> = (last_latest_slot..new_slot).collect();
-                // context for lock
-                {
-                    let mut lock = slots_task_queue.lock().await;
-                    for slot in new_block_slots {
-                        lock.push_back((slot, 0));
-                    }
+                let new_latest_slot = *new_block_slots.last().unwrap();
+
+                if latest_slot == new_latest_slot {
```
godmodegalactus, line 327: Force fetching of blocks, if slots were not updated for atleast 2 seconds.

godmodegalactus (changes requested):

src/workers/block_listenser.rs
aniketfuryrocks: Fixed in 94ad00d805b15b97950d227d35b2e26536b1da5f

src/workers/block_listenser.rs
aniketfuryrocks: Fixed in e61895092d12469746ed595a616e28a9dd507757

src/workers/block_listenser.rs
aniketfuryrocks: Fixed c5ef5a0cf72ccd7865944dbcd9fbda58729dd9a3

src/workers/block_listenser.rs
```diff
@@ -296,94 +294,55 @@ impl BlockListener {
...
+                        slot_que.split_off(slot_que.len().min(16))
+                    };
+
+                let index_futs = slots_to_get_blocks
+                    .iter()
+                    .map(|slot| self.index_slot(*slot, commitment_config, postgres.clone()));
+
+                for (index, status) in join_all(index_futs).await.into_iter().enumerate() {
+                    if status.is_err() {
+                        slot_que.push(slots_to_get_blocks[index]);
```
godmodegalactus, line 343: Should only try to get the slot 5 times. This will cause it to request slot infinite amount of time.

src/workers/block_listenser.rs
```diff
@@ -296,94 +294,55 @@ impl BlockListener {
...
+                let mut new_block_slots = (latest_slot..new_latest_slot).into_iter().collect();
+                latest_slot = new_latest_slot;
+
+                // need not queue up new slots if they are less than 16 and the queue is empty
+                let slots_to_get_blocks =
+                    if slot_que.is_empty() && (new_latest_slot - latest_slot) <= 16 {
+                        new_block_slots
+                    } else {
+                        slot_que.append(&mut new_block_slots);
+                        slot_que.split_off(slot_que.len().min(16))
```
godmodegalactus, line 334: instead of just taking first 16 slot you should do join all in batches of 16. If you just do it on 16 blocks and then loop back then there is a risk that if the queue increases and we never get the blocks at end of the queue.

src/workers/block_listenser.rs
```diff
@@ -296,94 +294,55 @@ impl BlockListener {
...
-                        lock.push_back((slot, 0));
+                let mut new_block_slots = (latest_slot..new_latest_slot).into_iter().collect();
+                latest_slot = new_latest_slot;
+
+                // need not queue up new slots if they are less than 16 and the queue is empty
+                let slots_to_get_blocks =
+                    if slot_que.is_empty() && (new_latest_slot - latest_slot) <= 16 {
+                        new_block_slots
+                    } else {
+                        slot_que.append(&mut new_block_slots);
```
godmodegalactus, line 333: new blocks should be added in the beginning of the queue.

godmodegalactus (changes requested):

src/workers/block_listenser.rs
aniketfuryrocks: Exactly. But max asked me to limit the number of calls to 16
https://discord.com/channels/791995070613159966/1022557374017441852/1077275278830743603

src/workers/block_listenser.rs
aniketfuryrocks: Won;t matter because we are splicing front the front

src/workers/block_listenser.rs
godmodegalactus: Hmm seems risky.

PR #64, opened 2023-02-22
use async channels

aniketfuryrocks:


src/workers/tx_sender.rs
```diff
@@ -72,107 +120,49 @@ impl TxSender {
...
-                        Ok(value) => match value {
-                            Some((sig, tx, slot)) => {
-                                sigs_and_slots.push((sig, slot));
-                                txs.push(tx);
-                            }
-                            None => {
-                                bail!("Channel Disconnected");
-                            }
-                        },
+                    match recv.try_recv() {
```
godmodegalactus, line 149: Please have the same logic as before with tokio::time::timeout etc.

godmodegalactus (changes requested):

src/workers/tx_sender.rs
aniketfuryrocks: Fixed

godmodegalactus (approved):

PR #66, opened 2023-02-22
Use less memory in BlockInformation #59 

aniketfuryrocks:


src/block_store.rs
```diff
@@ -26,14 +32,21 @@ impl BlockStore {
...
+                finalized_blockhash.clone(),
+                finalized_block,
+            ))),
             blocks: Arc::new({
                 let map = DashMap::new();
                 map.insert(confirmed_blockhash, confirmed_block);
                 map.insert(finalized_blockhash, finalized_block);
                 map
             }),
+            last_add_block_metric: Arc::new(RwLock::new(Instant::now())),
```
godmodegalactus, line 49: remove the add_block_metric logic from this PR. As you have introduced in another PR.

godmodegalactus (approved):

src/block_store.rs
aniketfuryrocks: Yeah this can be resolved during merging

PR #80, opened 2023-03-06
78 postgres batching

aniketfuryrocks:


src/workers/postgres.rs
```diff
@@ -183,32 +196,53 @@ impl Postgres {
...
+        Ok(self.session.read().await)
+    }
+
+    pub fn start(mut self, mut recv: PostgresMpscRecv) -> JoinHandle<anyhow::Result<()>> {
         tokio::spawn(async move {
             info!("Writing to postgres");
 
             while let Some(msg) = recv.recv().await {
-                MESSAGES_IN_POSTGRES_CHANNEL.dec();
+                MESSAGES_IN_POSTGRES_CHANNEL.inc();
```
godmodegalactus, line 238: decrement right?

src/workers/postgres.rs
```diff
@@ -162,15 +170,20 @@ impl Postgres {
...
-        self.client.read().await.execute(
-            r#"
-                INSERT INTO lite_rpc.Txs 
-                (signature, recent_slot, forwarded_slot, processed_slot, cu_consumed, cu_requested, quic_response)
-                VALUES
-                ($1, $2, $3, $4, $5, $6, $7)
-            "#,
-            &[&signature, &recent_slot, &forwarded_slot, &processed_slot, &cu_consumed, &cu_requested, &quic_response],
-        ).await?;
+        self.client
```
godmodegalactus, line 173: How is batching done ?

godmodegalactus (changes requested):

src/workers/postgres.rs
aniketfuryrocks: reverting

src/workers/postgres.rs
aniketfuryrocks: So that is taken care of internally by the library. U just have to pre prepare the client with the queries

godmodegalactus (approved):

src/workers/postgres.rs
godmodegalactus: There is no way to send multiple transactions at a time ?

godmodegalactus (approved):

src/workers/postgres.rs
mschneider: yes there is https://www.digitalocean.com/community/tutorials/sql-insert-multiple-rows it will dramatically speed up this code

src/workers/postgres.rs
```diff
@@ -96,35 +95,51 @@ impl Postgres {
...
 
-                bail!("Potsgres revival loop failed")
-            })
-        };
+        let update_tx_statement = client
+            .prepare(
+                r#"
+                UPDATE lite_rpc.txs 
+                SET processed_slot = $1, cu_consumed = $2, cu_requested = $3
+                WHERE signature = $4
```
mschneider, line 126: use multi-row insert with ON DUPLICATE KEY UPDATE so you can batch

PR #82, opened 2023-03-08
In block listner if there are errors while fetching slot we ignore an…

godmodegalactus:
…d continue


src/workers/block_listenser.rs
```diff
@@ -427,9 +429,17 @@ impl BlockListener {
...
             loop {
-                let new_slot = rpc_client
-                    .get_slot_with_commitment(commitment_config)
-                    .await?;
+                let new_slot = rpc_client.get_slot_with_commitment(commitment_config).await;
+
+                let new_slot = match new_slot {
+                    Ok(slot) => slot,
+                    Err(e) => {
+                        println!("error while fetching the slot {}", e.to_string());
```
aniketfuryrocks: Could u error! or warn! here instead of println!

aniketfuryrocks (approved):

PR #91, opened 2023-03-20
docker gso log

aniketfuryrocks:
- log gso
- log gso


Dockerfile
```diff
@@ -20,4 +20,5 @@ FROM debian:bullseye-slim as run
 RUN apt-get update && apt-get -y install ca-certificates libc6
 COPY --from=build /app/target/release/lite-rpc /usr/local/bin/
 
+CMD ethtool -k DEVICE | grep generic-segmentation-offload
```
godmodegalactus, line 23: DEVICE should be replace with the device name.
Add lines to enable it again and then repeat this line to check if it enabled.

godmodegalactus (changes requested):

PR #104, opened 2023-04-03
103 latest solanalite rpc upstream broken

aniketfuryrocks:
- upgrade deps
- upgrade deps manual
- use latest tpu client

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ members = [
 ]
 
 [workspace.dependencies]
-tokio = { version = "1.25.0", features = ["full"]}
+tokio = { version = "1.27.0", features = ["full"]}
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -9,39 +9,68 @@ members = [
     "bench"
 ]
 
+[workspace.dependencies]
+tokio = { version = "1.27.0", features = ["full"]}
+solana-sdk= { path = "../solana/sdk" }
```
godmodegalactus, line 14: should be pointing to a github repo like master  branch or crate.

godmodegalactus (changes requested):

Cargo.toml
aniketfuryrocks: Fixed

godmodegalactus:
no longer required

PR #106, opened 2023-04-04
store timestamps in sqlana

mschneider:
* record timestamp when sending transaction
* record timestamp when observing block on processed commitment
* record cluster time for each block


mschneider:
tried running test locally but couldn't figure out how to, github actions also seems broken 🙃

aniketfuryrocks:
> tried running test locally but couldn't figure out how to, github actions also seems broken 🙃

yeah. it's queued up in #104

godmodegalactus (approved):

PR #119, opened 2023-04-14
Optimize custom tpu 2

godmodegalactus:


bench/src/main.rs
```diff
@@ -39,17 +40,33 @@ async fn main() {
...
 
+    let join_res = join_all(tasks).await;
+
+    let mut run_num = 1;
+    for res in join_res {
+        match res {
+            Ok(metric) => {
+                info!("Run {run_num}: Sent and Confirmed {tx_count} tx(s) in {metric:?} with",);
+                // update avg metric
+                avg_metric += &metric;
```
aniketfuryrocks, line 60: This is not an accurate way top calculate avg.
cause join_all waits for all futures

src/workers/tpu_utils/tpu_connection_manager.rs
```diff
@@ -116,50 +121,162 @@ impl ActiveConnection {
...
-                already_connected,
-                endpoint.clone(),
-                addr,
-                exit_signal.clone(),
-            )
-            .await else {
-                return None;
-            };
+    // copied from solana code base
+    fn compute_receive_window_ratio_for_staked_node(
```
aniketfuryrocks, line 262: Is this function not exported from any crate from solana ?

src/workers/tpu_utils/tpu_connection_manager.rs
```diff
@@ -116,50 +121,162 @@ impl ActiveConnection {
...
-                    connection.open_uni(),
-                )
-                .await
-                {
-                    Ok(Ok(unistream)) => return Some(unistream),
-                    Ok(Err(_)) => (),
-                    Err(_) => return None,
+        last_stable_id: Arc<AtomicU64>,
+    ) {
+        for _ in 0..3 {
```
aniketfuryrocks, line 207: What's the 3 for. Can be made a constant

aniketfuryrocks (changes requested):

bench/src/main.rs
godmodegalactus: Ok i will update it later.

src/workers/tpu_utils/tpu_connection_manager.rs
godmodegalactus: It is not public but private function.

PR #122, opened 2023-04-16
Postgres test migration fix

aniketfuryrocks:


bench/src/metrics.rs
```diff
@@ -35,6 +35,9 @@ impl AddAssign<&Self> for Metric {
 
 impl DivAssign<u64> for Metric {
     fn div_assign(&mut self, rhs: u64) {
+        if rhs == 0 {
+            return;
+        }
```
mschneider, line 40: where is this used? i think panic is more appropriate here than return

src/workers/postgres.rs
```diff
@@ -23,36 +23,73 @@ use crate::encoding::BinaryEncoding;
 
 lazy_static::lazy_static! {
     pub static ref MESSAGES_IN_POSTGRES_CHANNEL: GenericGauge<prometheus::core::AtomicI64> = register_int_gauge!(opts!("literpc_messages_in_postgres", "Number of messages in postgres")).unwrap();
+    pub static ref POSTGRES_SESSION_ERRORS: GenericGauge<prometheus::core::AtomicI64> = register_int_gauge!(opts!("literpc_session_errors", "Number of failures while establishing postgres session")).unwrap();
+}
+
+const MAX_QUERY_SIZE: usize = 200_000; // 0.2 mb
+
+trait SchemaSize {
+    const DEFAULT_SIZE: usize = 0;
```
mschneider, line 32: can use 1 as a default

src/workers/postgres.rs
```diff
@@ -23,36 +23,73 @@ use crate::encoding::BinaryEncoding;
...
+    const MAX_SIZE: usize = 0;
+}
+
+const fn get_max_safe_inserts<T: SchemaSize>() -> usize {
+    let Some(n) = MAX_QUERY_SIZE.checked_div(T::DEFAULT_SIZE) else {
+        return 0;
```
mschneider, line 38: here as well

src/workers/postgres.rs
```diff
@@ -23,36 +23,75 @@ use crate::encoding::BinaryEncoding;
...
 pub struct PostgresTx {
-    pub signature: String,
-    pub recent_slot: i64,
-    pub forwarded_slot: i64,
-    pub forwarded_local_time: DateTime<Utc>,
+    pub signature: String,                   // 88 bytes
+    pub recent_slot: i64,                    // 8 bytes
+    pub forwarded_slot: i64,                 // 8 bytes
+    pub forwarded_local_time: DateTime<Utc>, // 8 bytes
     pub processed_slot: Option<i64>,
```
godmodegalactus, line 58: Option should be 1 bytes or 9 bytes right?

src/workers/postgres.rs
```diff
@@ -327,37 +388,59 @@ impl Postgres {
                     }
                 }
 
-                let Ok(session) = self.get_session().await else {
+                if tx_batch.is_empty() && block_batch.is_empty() && update_batch.is_empty() {
+                    tokio::time::sleep(Duration::from_millis(400)).await;
```
godmodegalactus, line 392: add continue over here.

src/workers/postgres.rs
```diff
@@ -299,23 +343,40 @@ impl Postgres {
...
+
+            let mut tx_batch: Vec<PostgresTx> = Vec::with_capacity(TX_MAX_CAPACITY);
+            let mut block_batch: Vec<PostgresBlock> = Vec::with_capacity(BLOCK_MAX_CAPACITY);
+            let mut update_batch =
+                Vec::<(PostgresUpdateTx, String)>::with_capacity(UPDATE_MAX_CAPACITY);
+
+            let mut session_establish_error = false;
 
             loop {
                 loop {
```
godmodegalactus, line 358: Could move all the vectors in the loop so that they are cleared at the end of the loop.

godmodegalactus (changes requested):

src/workers/postgres.rs
aniketfuryrocks: It is not sent in the query. I am not calculating struct size, but the size of the actual query

src/workers/postgres.rs
aniketfuryrocks: oh yeah good catch

src/workers/postgres.rs
aniketfuryrocks: no we don't want reallocation of memory. After a fixed capacity is allocated, just clear it and re use it. More efficient.

src/workers/postgres.rs
aniketfuryrocks: done

bench/src/metrics.rs
aniketfuryrocks: avg metric. if there were no runs then bench exits with a log

src/workers/postgres.rs
```diff
@@ -327,37 +388,60 @@ impl Postgres {
                     }
                 }
 
-                let Ok(session) = self.get_session().await else {
+                if tx_batch.is_empty() && block_batch.is_empty() && update_batch.is_empty() {
+                    tokio::time::sleep(Duration::from_millis(400)).await;
```
mschneider, line 392: lower sleep here to 50ms, no need to wait so long

aniketfuryrocks: fixed

PR #124, opened 2023-04-18
Adding transaction replay service

godmodegalactus:


src/bridge.rs
```diff
@@ -254,6 +270,19 @@ impl LiteRpcServer for LiteBridge {
                 e
             );
         }
+
+        if let Some(tx_replay_sender) = &self.tx_replay_sender {
+            let max_replay = max_retries.map_or(10, |x| x) as usize;
```
mschneider, line 275: should be a cli arg

src/bridge.rs
```diff
@@ -254,6 +270,19 @@ impl LiteRpcServer for LiteBridge {
...
+            let replay_at = Instant::now() + RETRY_AFTER;
```
mschneider, line 276: cli / config

PR #126, opened 2023-04-19
upgrade jsonrpsee and fixed 1 bug of early return

aniketfuryrocks:


src/workers/block_listenser.rs
```diff
@@ -331,17 +331,23 @@ impl BlockListener {
...
             {
                 // none if transaction succeeded
-                sink.send(&RpcResponse {
-                    context: RpcResponseContext {
-                        slot,
-                        api_version: None,
-                    },
-                    value: serde_json::json!({ "err": err }),
-                })?;
+                let _res = sink
```
aniketfuryrocks, line 338: Bug fix. Should not return using `?`

godmodegalactus: Yes true

godmodegalactus (approved):

PR #133, opened 2023-04-29
rm fin_block and conf_block_recv

aniketfuryrocks:
We don't require these metrics.
They can be acheived using the tt_recv metrics count.

### DIFF
--- a/src/workers/block_listenser.rs
+++ b/src/workers/block_listenser.rs
@@ -55,10 +55,6 @@ lazy_static::lazy_static! {
         "Time to receive finalized block from block subscribe",
     ))
     .unwrap();
-    static ref FIN_BLOCKS_RECV: IntCounter =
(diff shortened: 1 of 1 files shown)


src/workers/block_listenser.rs
```diff
@@ -55,10 +55,6 @@ lazy_static::lazy_static! {
         "Time to receive finalized block from block subscribe",
     ))
     .unwrap();
-    static ref FIN_BLOCKS_RECV: IntCounter =
```
godmodegalactus, line 58: why do you want to remove this? Could be interesting later to debug.

aniketfuryrocks: The `TT_RECV_FIN_BLOCK` and `TT_RECV_FIN_BLOCK` already has a count property in them.
So this is the same thing done twice here

godmodegalactus (approved):

PR #136, opened 2023-05-02
save transaction in a seperate csv

godmodegalactus:


aniketfuryrocks:
Instead shouldn't we open up the transaction history as a custom json_rpc call ?

godmodegalactus:
> Instead shouldn't we open up the transaction history as a custom json_rpc call ? A simple bash could poll it out

We cannot get how much how much time was taken to confirm the transaction on client side.

PR #137, opened 2023-05-02
Block Listner Channel Fix

aniketfuryrocks:


src/workers/block_listenser.rs
```diff
@@ -437,10 +428,18 @@ impl BlockListener {
                         let retry_at = tokio::time::Instant::now()
                             .checked_add(Duration::from_millis(10))
                             .unwrap();
-                        let _ = slot_retry_queue_sx.send((slot, retry_at));
+
+                        slot_retry_queue_sx.send((slot, retry_at))?;
```
aniketfuryrocks, line 432: It should error out exit if the channel is closed

src/workers/block_listenser.rs
```diff
@@ -402,26 +402,17 @@ impl BlockListener {
         estimated_slot: Arc<AtomicU64>,
     ) -> JoinHandle<anyhow::Result<()>> {
         let (slot_retry_queue_sx, mut slot_retry_queue_rx) = tokio::sync::mpsc::unbounded_channel();
-        let (block_schedule_queue_sx, block_schedule_queue_rx) =
-            async_channel::unbounded::<Slot>();
+        let (block_schedule_queue_sx, block_schedule_queue_rx) = async_channel::unbounded::<Slot>();
 
         // task to fetch blocks
-        for _i in 0..8 {
+        for _ in 0..8 {
```
aniketfuryrocks, line 408: I think we should select all the tasks we create.

mschneider (approved):

PR #144, opened 2023-06-06
Creating a new package for lite-rpc algorithms

godmodegalactus:


mschneider:
this is very misleading package structure, the code in the algorithms package is not really algorithmic, it's 99% io code. 
recommend to start with factoring out services and binaries and leave the rest in a core package.

godmodegalactus:
Created a new library called solana-lite-rpc-core and solana-lite-rpc-services

PR #147, opened 2023-06-12
migrate

aniketfuryrocks:
- block_processor and bridge
- fmt
- tpu service

### DIFF
--- a/services/src/tpu_utils/tpu_service.rs
+++ b/services/src/tpu_utils/tpu_service.rs
@@ -1,13 +1,14 @@
-use anyhow::Result;
+use anyhow::bail;
 use dashmap::DashMap;
 use log::{error, info};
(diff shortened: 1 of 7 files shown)


services/src/tpu_utils/tpu_service.rs
```diff
@@ -177,15 +178,23 @@ impl TpuService {
...
+                &self.rpc_ws_address,
+                update_slot,
+            )
+            .await else {
+                bail!("current slot fetch task exited");
+            };
+
+            error!("slot fetch task error: {err:?}");
+
+            tokio::time::sleep(Duration::from_millis(2000)).await;
```
godmodegalactus, line 193: We wait for slot to be polled for 2s and then again wait here for 2s so in total 4s whereas before we use to wait only for 2s. So remove this wait.

godmodegalactus (changes requested):

PR #150, opened 2023-06-16
counters

aniketfuryrocks:


lite-rpc/src/main.rs
```diff
@@ -31,6 +31,11 @@ async fn get_identity_keypair(identity_from_cli: &String) -> Keypair {
     }
 }
 
+lazy_static::lazy_static! {
+    static ref RESTARTS: IntCounter =
+    register_int_counter!(opts!("literpc_rpc_restarts", "Nutber of times lite rpc restarted")).unwrap();
```
godmodegalactus, line 36: nit! nutber

godmodegalactus (changes requested):

core/src/solana_utils.rs
```diff
@@ -61,7 +61,7 @@ impl SolanaUtils {
     }
 
     pub async fn poll_slots(
-        rpc_client: Arc<RpcClient>,
```
godmodegalactus, line 64: Why did you remove arc?

aniketfuryrocks: For the lifetime and type of function. It doesn't require the ownership of rpc_client. Therefore the reference of the inner of an Atomic Reference Counter.

lite-rpc/src/main.rs
```diff
@@ -79,13 +75,52 @@ pub async fn main() -> anyhow::Result<()> {
...
+        || if let Ok(enable_postgres_env_var) = env::var("PG_ENABLED") {
+            enable_postgres_env_var != "false"
+        } else {
+            false
+        };
+
+    args
+}
+
+#[tokio::main]
```
godmodegalactus, line 97: I prefer it as before specifying 16 worker threads.

services/src/tpu_utils/tpu_service.rs
```diff
@@ -163,48 +163,42 @@ impl TpuService {
...
-                    );
-                    break;
-                }
-            } else {
-                nb_errror = 0;
+                log::info!("Got error while polling slot {}", err);
             }
-        }
+
+            bail!("Reached max amount of errors to fetch latest slot, exiting poll slot loop")
```
godmodegalactus, line 196: Should also log an error! here

services/src/transaction_replayer.rs
```diff
@@ -45,56 +42,44 @@ impl TransactionReplayer {
...
+                        error!("error while scheduling replay ({})", e);
+                        continue;
+                    } else {
+                        MESSAGES_IN_REPLAY_QUEUE.inc();
                     }
                 }
             }
-            Ok(())
+
+            bail!("transaction replay channel broken");
```
godmodegalactus, line 82: here to add an explicit error! log

godmodegalactus (changes requested):

services/src/transaction_replayer.rs
aniketfuryrocks: same here. It will be logged down the line

services/src/tpu_utils/tpu_service.rs
godmodegalactus: You are sure right?

godmodegalactus: I prefer it having it twice rather than none because something went wrong with tokio stuff. Lite-rpc had problems in past where it went down without showing any logs which service went down.

godmodegalactus:
Where is the prometheus variable that is 1 when RPC is replying and 0 when it is not?
Similarly, for postgres?

aniketfuryrocks:
> 

It will restart if any of the 2 fail or any service fails

lite-rpc/src/main.rs
```diff
@@ -79,13 +75,53 @@ pub async fn main() -> anyhow::Result<()> {
...
+    let args = get_args();
+
+    let main: AnyhowJoinHandle = tokio::spawn(async move {
+        loop {
+            let Err(err) = start_lite_rpc(args.clone()).await else{
+                bail!("LiteBridge services returned without error");
+            };
+
+            // log and restart
+            log::error!("Services quit unexpectedly {err:?} restarting in {RESTART_DURATION:?}");
```
aniketfuryrocks, line 110: It will log error here

godmodegalactus: Have you tested if it is working?

lite-rpc/src/bridge.rs
```diff
@@ -57,6 +57,18 @@ lazy_static::lazy_static! {
     register_int_counter!(opts!("literpc_rpc_airdrop", "RPC call to request airdrop")).unwrap();
     static ref RPC_SIGNATURE_SUBSCRIBE: IntCounter =
     register_int_counter!(opts!("literpc_rpc_signature_subscribe", "RPC call to subscribe to signature")).unwrap();
+    static ref WS_SERVER_FAIL: IntCounter =
+    register_int_counter!(opts!("literpc_rpc_ws_server_fail", "WebSocket server failed")).unwrap();
+    static ref METRICS_SERVICE_FAIL: IntCounter =
+    register_int_counter!(opts!("literpc_rpc_metrics_service_fail", "Metrics service failed")).unwrap();
+    static ref HTTP_SERVER_FAIL: IntCounter =
+    register_int_counter!(opts!("literpc_rpc_http_server_fail", "Http server failed")).unwrap();
+    static ref PROMETHEUS_SERVER_FAIL: IntCounter =
```
aniketfuryrocks, line 66: Prometheus fail

godmodegalactus:
> > 
> 
> It will restart if any of the 2 fail or any service fails. Log error and increment the restart counter. We can set alert when the restart counter increases.

It does not work. Simple we want a gauge that will go to 0 when triton rpc is working and will go to 0 if is not working. And it should again go to one when it is working again. So that we can display on dashboard that rpc is down or postgres is down etc. If you have questions discuss with Pan. Have you discussed with Pan about this idea and has he give a green light?

lite-rpc/src/main.rs
aniketfuryrocks: Yes. Earlier it didn't log cause it was bailing out from the main. Which for some reason is not captured by Prometheus.

godmodegalactus: Ok, lets see then on fly if it works.

godmodegalactus (approved):

PR #166, opened 2023-08-18
Implementation of getBlocks and implementation of various block storage

godmodegalactus:
Implementing getBlocks,
Replacing block_store to block_information_store.
Implementing inmemory block store.
Implementing mixed_strategy to fetch block, For few blocks fetch from in memory, for more than that either from persistent block storage or from faithful.

### DIFF
--- a/lite-rpc/src/bridge.rs
+++ b/lite-rpc/src/bridge.rs
@@ -17,7 +17,10 @@ use solana_lite_rpc_core::{
 };
 use solana_rpc_client::nonblocking::rpc_client::RpcClient;
 use solana_rpc_client_api::{
-    config::{RpcContextConfig, RpcRequestAirdropConfig, RpcSignatureStatusConfig},
+    config::{
+        RpcBlockConfig, RpcContextConfig, RpcEncodingConfigWrapper, RpcRequestAirdropConfig,
+        RpcSignatureStatusConfig,
+    },
     response::{Response as RpcResponse, RpcBlockhash, RpcResponseContext, RpcVersionInfo},
 };
 use solana_sdk::{
@@ -332,5 +335,19 @@ impl LiteRpcServer for LiteBridge {
         );
 
         Ok(())
+    }
+
+    async fn get_block(
+        &self,
+        slot: u64,
+        _config: Option<RpcEncodingConfigWrapper<RpcBlockConfig>>,
+    ) -> crate::rpc::Result<Option<UiConfirmedBlock>> {
+        if let Some(block_storage) = &self.block_storage {
+            Ok(block_storage.get(slot).await)
+        } else {
+            Err(jsonrpsee::core::Error::Custom(
+                "Does not support get_block".to_string(),
+            ))
+        }
     }
 }
--- a/storage/src/multiple_strategy_block_store.rs
+++ b/storage/src/multiple_strategy_block_store.rs
@@ -1,0 +1,123 @@
+// A mixed block store,
+// Stores confirmed blocks in memory
+// Finalized blocks in long term storage of your choice
+// Fetches legacy blocks from faithful
+
+use crate::block_stores::inmemory_block_store::InmemoryBlockStore;
+use async_trait::async_trait;
+use solana_lite_rpc_core::block_storage_interface::{BlockStorageImpl, BlockStorageInterface};
+use solana_rpc_client::nonblocking::rpc_client::RpcClient;
+use solana_rpc_client_api::config::RpcBlockConfig;
+use solana_sdk::{
+    commitment_config::{CommitmentConfig, CommitmentLevel},
+    slot_history::Slot,
+};
+use solana_transaction_status::{TransactionDetails, UiConfirmedBlock, UiTransactionEncoding};
+use std::sync::{
+    atomic::{AtomicU64, Ordering},
+    Arc,
+};
+
+pub struct MultipleStrategyBlockStorage {
+    inmemory_for_storage: InmemoryBlockStore, // for confirmed blocks
+    persistent_block_storage: BlockStorageImpl, // for persistent block storage
+    faithful_rpc_client: Arc<RpcClient>,      // to fetch legacy blocks from faithful
+    number_of_slots_in_memory: usize,
+    last_confirmed_slot: Arc<AtomicU64>,
+}
+
+impl MultipleStrategyBlockStorage {
+    pub fn new(
+        persistent_block_storage: BlockStorageImpl,
+        faithful_url: String,
+        number_of_slots_in_memory: usize,
+    ) -> Self {
+        Self {
+            inmemory_for_storage: InmemoryBlockStore::new(number_of_slots_in_memory),
+            persistent_block_storage,
+            faithful_rpc_client: Arc::new(RpcClient::new(faithful_url)),
+            number_of_slots_in_memory,
+            last_confirmed_slot: Arc::new(AtomicU64::new(0)),
+        }
+    }
+
+    pub async fn get_in_memory_block(&self, slot: Slot) -> Option<UiConfirmedBlock> {
+        self.inmemory_for_storage.get(slot).await
+    }
+}
+
+#[async_trait]
+impl BlockStorageInterface for MultipleStrategyBlockStorage {
+    async fn save(
+        &self,
+        slot: solana_sdk::slot_history::Slot,
+        block: solana_transaction_status::UiConfirmedBlock,
+        commitment: solana_sdk::commitment_config::CommitmentLevel,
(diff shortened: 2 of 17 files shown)


core/src/block_processor.rs
```diff
@@ -80,6 +82,11 @@ impl BlockProcessor {
             .await
             .context("failed to get block")?;
 
+        // save the block in storage
+        if let Some(block_store) = &self.block_store {
+            block_store.save(slot, block.clone(), commitment_config.commitment).await;
```
musitdev: I think the clone of the full block can be avoided because the rest of the process use block data but doesn't own them.

lite-rpc/src/bridge.rs
```diff
@@ -65,10 +70,11 @@ lazy_static::lazy_static! {
 /// A bridge between clients and tpu
 pub struct LiteBridge {
     pub rpc_client: Arc<RpcClient>,
+    pub faithful_rpc_client: Option<Arc<RpcClient>>,
```
musitdev, line 73: faithfull service use Solana RPC client to get the data?

core/src/block_processor.rs
```diff
@@ -80,6 +82,11 @@ impl BlockProcessor {
             .await
             .context("failed to get block")?;
 
+        // save the block in storage
+        if let Some(block_store) = &self.block_store {
```
musitdev, line 93: I think you should call it block_storage instead of block_store because it creates some confusion. block_store is already use elsewhere for other thing; See line 101 for example.

core/src/block_processor.rs
```diff
@@ -91,7 +98,7 @@ impl BlockProcessor {
         let blockhash = block.blockhash;
         let parent_slot = block.parent_slot;
 
-        if let Some(block_store) = &self.block_store {
+        if let Some(block_store) = &self.block_information_store {
```
musitdev, line 110: Another block_store.

lite-rpc/src/bridge.rs
```diff
@@ -467,6 +493,37 @@ impl LiteRpcServer for LiteBridge {
 
         Ok(())
     }
+
+    async fn get_block(
+        &self,
+        slot: u64,
+        _config: Option<RpcEncodingConfigWrapper<RpcBlockConfig>>,
+    ) -> crate::rpc::Result<Option<UiConfirmedBlock>> {
+        if let Some(faithful_rpc_client) = &self.faithful_rpc_client {
```
musitdev, line 502: getBlock only query the faithfull plugin not the storage?

storage/src/multiple_strategy_block_store.rs
```diff
@@ -0,0 +1,109 @@
...
+        block: solana_transaction_status::UiConfirmedBlock,
+        commitment: solana_sdk::commitment_config::CommitmentLevel,
+    ) {
+        if commitment == CommitmentLevel::Confirmed {
+            self.inmemory_for_storage
+                .save(slot, block, commitment)
+                .await;
+        } else if commitment == CommitmentLevel::Finalized {
+            let block_in_mem = self.inmemory_for_storage.get(slot).await;
+            if block_in_mem.is_none() {
```
musitdev, line 63: You should add the contains method to the Inmemory storage and use it to define if the block exists.
Otherwise, you can write it:
```
if let None = self.inmemory_for_storage.get(slot).await {...}
```

storage/src/multiple_strategy_block_store.rs
```diff
@@ -0,0 +1,109 @@
...
+        &self,
+        slot: solana_sdk::slot_history::Slot,
+    ) -> Option<solana_transaction_status::UiConfirmedBlock> {
+        let last_confirmed_slot = self.last_confirmed_slot.load(Ordering::Relaxed);
+        if slot > last_confirmed_slot {
+            None
+        } else {
+            let diff = last_confirmed_slot.saturating_sub(slot);
+            if diff < self.number_of_slots_in_memory as u64 {
+                let block = self.inmemory_for_storage.get(slot).await;
```
musitdev, line 95: Here the same you can write it:
```
if let Some(block) = self.inmemory_for_storage.get(slot).await {
 return block;
}
```
ot use contains.

storage/src/multiple_strategy_block_store.rs
```diff
@@ -0,0 +1,109 @@
...
+use async_trait::async_trait;
+use solana_lite_rpc_core::block_storage_interface::{BlockStorageInterface, BlockStorageImpl};
+use solana_rpc_client::nonblocking::rpc_client::RpcClient;
+use solana_rpc_client_api::config::RpcBlockConfig;
+use solana_sdk::{commitment_config::{CommitmentConfig, CommitmentLevel}, slot_history::Slot};
+use solana_transaction_status::{TransactionDetails, UiTransactionEncoding, UiConfirmedBlock};
+use std::sync::{Arc, atomic::{AtomicU64, Ordering}};
+
+pub struct MultipleStrategyBlockStorage {
+    inmemory_for_storage: InmemoryBlockStore, // for confirmed blocks
```
musitdev, line 22: In the code, the inmemory_for_storage act has a cache for the persistent_block_storage. I think the persittent_block_storage cache should be implemented in the persistent storage. Usually db has already some cache that can make it useless. You can use LRUCache that can be easier to use and more efficient.

storage/src/multiple_strategy_block_store.rs
```diff
@@ -0,0 +1,109 @@
...
+            let diff = last_confirmed_slot.saturating_sub(slot);
+            if diff < self.number_of_slots_in_memory as u64 {
+                let block = self.inmemory_for_storage.get(slot).await;
+                if block.is_some() {
+                    return block;
+                }
+            }
+            // TODO: Define what data is expected that is definetly not in persistant block storage like data after epoch - 1
+            // check persistant block
+            let persistant_block = self.persistent_block_storage.get(slot).await;
```
musitdev, line 102: The same.

lite-rpc/src/bridge.rs
godmodegalactus: Yes It should work with current RPCClient, any issues will open PR on faithful side. Recently I had created a fix for the same issue on faithful side.

storage/src/multiple_strategy_block_store.rs
godmodegalactus: The idea of inmemory storage is to store the confirmed block and not finalized blocks. So when blocks are finalized they will be stored persistent storage. In anyways usually most of the clients would ask for recent blocks so some kind of inmemory persistence will be useful.

core/src/block_processor.rs
godmodegalactus: Done

core/src/block_processor.rs
godmodegalactus: Done

lite-rpc/src/bridge.rs
musitdev: Ok, I didn't know. But they don't implement all RPCclient API call?

lite-rpc/src/bridge.rs
musitdev: Ok fine

storage/src/multiple_strategy_block_store.rs
musitdev: Ok, it wasn't in my version. I think you should use match instead of if none else if let some. It's easier to read.

storage/src/multiple_strategy_block_store.rs
musitdev: Ok, you should use the match because you process all case (some and none)

storage/src/multiple_strategy_block_store.rs
musitdev: As I understand, blocks have 2 states confirmed and finalized. Does all confirmed block become finalized?
If yes, you can use a LRUCache and add all block to it. By setting the size to 20 I think it's enough to cache all confirmed block. You add all block to the cache and only finalized are written to the DB. This way you always get the confirmed first if the block is in the cache, and you don't have to manage the removing.
If not, the LRUCache will remove them after 20 blocks, so they don't need to be removed.
In my opinion, you don't have to define the interface BlockStorageInterface because the impl don't have the same use case. InmemoryBlockStore is a cache, so it needs in addition: contains and remove (auto remove is not always done by the cache), MultipleStrategyBlockStorage need get and save only and persistent is hidden behind MultipleStrategyBlockStorage. So in the end, only MultipleStrategyBlockStorage should be public.
FYI, I already use this cache: https://crates.io/crates/lru

musitdev (commented):
I add some comment about LRUcache use.

PR #169, opened 2023-08-21
Quic Proxy: cleanup/consolidate tests, handle flaky tests, add explicit connection close

grooviegermanikus:
- add explicit tpu connect shutdown
- remove redundant integration test (literpc_tpu_quic_server_integrationtest.rs) and use the quic_proxy_tpu_integrationtest.rs for everything
- remove some commands from test.bash


test.bash
```diff
@@ -19,8 +15,8 @@ sleep 20 && solana airdrop 10000
 echo "Starting LiteRpc"
 cargo run &
 
-echo "Running cargo tests in 20s"
-sleep 20 && cargo test
```
godmodegalactus, line 23: I think there are other tests that will not be done if you remove cargo tests

godmodegalactus (changes requested):

test.bash
grooviegermanikus: hm don't think so. cargo test got executed twice in the whole process.
anyway - will revert this - it's not relevant for the PR

godmodegalactus (dismissed):

godmodegalactus (dismissed):

aniketfuryrocks (approved):

aniketfuryrocks (approved):

godmodegalactus (approved):

PR #171, opened 2023-08-21
tpu agends must handle connection shutdown explicitly

grooviegermanikus:


grooviegermanikus:
will integrate this change with https://github.com/blockworks-foundation/lite-rpc/pull/169 and close this PR

PR #173, opened 2023-08-28
Base interface def

musitdev:
This is an example of the RPC v2 architecture implementation for the Steam part.
An all-in-one process implementation is done in `allinonerpc` crate that implements the stream logic for in process.
A module is defined for each domain. Each domain get a Stream or Sink (depending on if it's a produced or a consumer) and process them.
I put some example for Block, BlockInfo, Slot and Epoch in each module depending on the stream defined in the spec: https://github.com/blockworks-foundation/lite-rpc/blob/rpc_doc/docs/architecture.md.
It can help to understand the main stream architecture.
The RPC part is missing. Each module implement its own RPC entry point.

### DIFF
--- a/allinonerpc/src/main.rs
+++ b/allinonerpc/src/main.rs
@@ -1,4 +1,5 @@
 use crate::stream::BlockInfokStreamConnector;
+use crate::stream::EpochStreamConnector;
 use crate::stream::FullblockStreamConnector;
 use crate::stream::SlotStreamConnector;
 use tokio_stream::StreamExt;
@@ -26,13 +27,21 @@ pub async fn main() -> anyhow::Result<()> {
     let block_sink = block_stream.get_sender_sink();
     let block_info_stream = BlockInfokStreamConnector::new();
     let block_info_sink = block_info_stream.get_sender_sink();
-    let history_handle = history::start_module(block_stream.subscribe()).await;
-    let send_tx_handle = sendtx::start_module(block_info_stream.subscribe()).await;
+    let epoch_stream = EpochStreamConnector::new();
+    let epoch_sink = epoch_stream.get_sender_sink();
+    let history_handle =
+        history::start_module(block_stream.subscribe(), epoch_stream.subscribe()).await;
+    let send_tx_handle = sendtx::start_module(
+        block_info_stream.subscribe(),
+        slot_connector.subscribe_slot(),
+    )
+    .await;
     let consensus_handle = consensus::start_module(
         GRPC_URL.to_string(),
         slot_sender,
         block_sink,
         block_info_sink,
+        epoch_sink,
     )
     .await;
 
--- a/allinonerpc/src/main.rs
+++ b/allinonerpc/src/main.rs
@@ -1,0 +1,54 @@
+use crate::stream::BlockInfokStreamConnector;
+use crate::stream::FullblockStreamConnector;
+use crate::stream::SlotStreamConnector;
+use tokio_stream::StreamExt;
+
+mod stream;
+
+const GRPC_URL: &str = "http://127.0.0.0:10000";
+
+#[tokio::main(flavor = "multi_thread", worker_threads = 8)]
+pub async fn main() -> anyhow::Result<()> {
+    //slot connector
+    let slot_connector = SlotStreamConnector::new(100);
+    let slot_notifier = slot_connector.get_sender();
+    //convert broadcast to a sink
+    let (slot_sender, mut slot_receiver) = futures::channel::mpsc::unbounded::<u64>();
+    tokio::spawn(async move {
+        while let Some(slot) = slot_receiver.next().await {
+            if let Err(err) = slot_notifier.send(slot) {
+                log::error!("Error during slot broadcast from sink:{err}");
+            }
+        }
+    });
+
+    let block_stream = FullblockStreamConnector::new();
+    let block_sink = block_stream.get_sender_sink();
+    let block_info_stream = BlockInfokStreamConnector::new();
+    let block_info_sink = block_info_stream.get_sender_sink();
+    let history_handle = history::start_module(block_stream.subscribe()).await;
+    let send_tx_handle = sendtx::start_module(block_info_stream.subscribe()).await;
+    let consensus_handle = consensus::start_module(
+        GRPC_URL.to_string(),
+        slot_sender,
+        block_sink,
+        block_info_sink,
+    )
+    .await;
+
+    tokio::select! {
+        err = history_handle => {
+            // This should never happen
+            unreachable!("{err:?}")
+        }
+        err = consensus_handle => {
+            // This should never happen
+            unreachable!("{err:?}")
+        }
+        err = send_tx_handle => {
+            // This should never happen
+            unreachable!("{err:?}")
+        }
+
+    }
+}
--- a/allinonerpc/src/stream.rs
+++ b/allinonerpc/src/stream.rs
@@ -1,0 +1,117 @@
+use futures::channel::mpsc::SendError as FuturesSendError;
+use futures::{channel::mpsc, sink::Sink, stream::Stream};
+use solana_lite_rpc_core::block_store::BlockInformation;
+use solana_sdk::epoch_info::EpochInfo;
+use solana_transaction_status::EncodedConfirmedBlock;
+use thiserror::Error;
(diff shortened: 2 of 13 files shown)


modules/history/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+    let handle = tokio::spawn(async move {
+        pin!(full_block_stream);
+        pin!(epoch_steam);
+
+        loop {
+            tokio::select! {
+                Some(full_block) = full_block_stream.next() => {
+                    //process full block
+                }
+                Some(epoch) = epoch_steam.next() => {
```
godmodegalactus, line 21: What is epoch stream?

godmodegalactus (changes requested):

allinonerpc/src/stream.rs
```diff
@@ -0,0 +1,117 @@
...
+    #[error("Error during during send slot '{0}'")]
+    SendSlotError(String),
+}
+
+pub type Slot = u64;
+
+macro_rules! create_stream {
+    ($connector_name:ident,$connector_sender:ident, $stream_data:ident) => {
+        pub struct $connector_name {
+            data_receiver: mpsc::UnboundedReceiver<$stream_data>,
```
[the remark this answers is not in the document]
musitdev: Yes, most of the communication is 1 to 1. The API is done in a way that only one receiver can be created. For broadcast, I put an example with slot subscription.
I update the architecture diagram to show the REST API per module and the message exchanged between them: https://github.com/blockworks-foundation/lite-rpc/blob/rpc_doc/docs/architecture.md#summary-diagram

modules/consensus/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    #[error("IO Error during history processing '{0}'")]
+    IOError(#[from] std::io::Error),
+
+    #[error("Geyser Error during history processing '{0}'")]
+    GeyserError(#[from] GeyserGrpcClientError),
+}
+
+pub async fn start_module(
+    grpc_url: String,
+    slot_sinck: impl Sink<u64, Error = FuturesSendError> + std::marker::Send + 'static,
```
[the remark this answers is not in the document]
musitdev: Yes I correct

modules/history/src/lib.rs
musitdev: It sends new epoch info at epoch change. History need to know when an epoch change to update its storage.

allinonerpc/src/stream.rs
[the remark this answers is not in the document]
godmodegalactus: I think it is better to assume that every data will be required by multiple consumers. We will have to add a mpsc module and then a broadcast module anyways for each type of data in consenseus.

allinonerpc/src/stream.rs
```diff
@@ -0,0 +1,117 @@
...
+                    data_sender,
+                }
+            }
+
+            pub fn subscribe(self) -> impl Stream<Item = $stream_data> {
+                self.data_receiver
+            }
+
+            //return an error if the fullblock subscription hasn't been done.
+            pub fn get_sender_sink(&self) -> impl Sink<$stream_data, Error = FuturesSendError> {
```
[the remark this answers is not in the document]
godmodegalactus: Normally each consumer should get a block only once, with ,multiple senders could risk of sending same data multiple times. This aggregation should be done before sending on the channel. I do not see a usecase where we need multiple senders.

modules/consensus/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+            SubscribeRequestFilterBlocks {
+                account_include: vec![], //vec![SYSTEM_PROGRAM.to_string()],
+                include_transactions: Some(true),
+                include_accounts: None,
+                include_entries: None,
+            },
+        );
+
+        let mut client = GeyserGrpcClient::connect(grpc_url, None::<&'static str>, None)?;
+        let mut confirmed_stream = client
```
grooviegermanikus, line 76: although coupling by technology (geyser) makes some sense here I wonder if we should isolate the different sources on a per-type-basis a much as possible - i.e. one service/submodule for slots, one for blocks, one for accounts

modules/consensus/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+use tokio::task::JoinHandle;
+use yellowstone_grpc_client::GeyserGrpcClient;
+use yellowstone_grpc_client::GeyserGrpcClientError;
+use yellowstone_grpc_proto::geyser::CommitmentLevel;
+use yellowstone_grpc_proto::prelude::SubscribeRequestFilterAccounts;
+use yellowstone_grpc_proto::prelude::SubscribeRequestFilterBlocks;
+use yellowstone_grpc_proto::prelude::{subscribe_update::UpdateOneof, SubscribeRequestFilterSlots};
+
+#[derive(Debug, Error)]
+pub enum HistoryError {
```
grooviegermanikus, line 20: IMO that type should go into _history_ module

allinonerpc/src/main.rs
```diff
@@ -0,0 +1,63 @@
...
+use crate::stream::FullblockStreamConnector;
+use crate::stream::SlotStreamConnector;
+use tokio_stream::StreamExt;
+
+mod stream;
+
+const GRPC_URL: &str = "http://127.0.0.0:10000";
+
+#[tokio::main(flavor = "multi_thread", worker_threads = 8)]
+pub async fn main() -> anyhow::Result<()> {
```
grooviegermanikus, line 12: I like the idea of a centralized wiring + dependency injection;
we maybe should have multiple flavors of that (e.g.lite-rpc only for send_tx + tx confirm, no history, no full blocks) early to see if the architecture is decoupled enough to support that

modules/consensus/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+use futures::SinkExt;
+use futures_util::StreamExt;
+use solana_lite_rpc_core::block_store::BlockInformation;
+use solana_sdk::epoch_info::EpochInfo;
+use solana_transaction_status::EncodedConfirmedBlock;
+use std::collections::HashMap;
+use thiserror::Error;
+use tokio::pin;
+use tokio::task::JoinHandle;
+use yellowstone_grpc_client::GeyserGrpcClient;
```
grooviegermanikus, line 12: IMO we should move technology/infrastructure code for getting/sending data out of the domain modules into infrastructure-modules (e.g. infrastructure/geyser-subscription-yellowstone)

modules/sendtx/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+use tokio::pin;
+use tokio::task::JoinHandle;
+use tokio_stream::wrappers::errors::BroadcastStreamRecvError;
+
+pub async fn start_module(
+    block_info_steam: impl Stream<Item = BlockInformation> + std::marker::Send + 'static,
+    slot_steam: impl Stream<Item = Result<u64, BroadcastStreamRecvError>> + std::marker::Send + 'static,
+) -> JoinHandle<std::io::Result<()>> {
+    let handle = tokio::spawn(async move {
+        pin!(slot_steam);
```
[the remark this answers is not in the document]
musitdev: It's the main loop that process input stream data. The select wait for data from the 2 stream. When a data is available, it is processed. The process should be very short or spawn in a specific task. This pattern can be used to avoid to use mutex of thread safe share structure as you can process share structure in one task and use an adapted algorithm because all data access are done in this loop.. 
If you're interested this is a more complex (and very raw) implementation that manage geyser account,-slot-epoch-block, bootstrap, RPC access and leader schedule and dump processing: https://github.com/blockworks-foundation/solana-rpc-v2/blob/0efb26c01b6c02013a74bf6aa72072564a95dd5a/stake_aggregate/src/main.rs#L206
I've to refactor everything to make is more readable, but the idea is here. As you can see, there's no lock or mutex to share the data, and I'm always sure that no task can block another. For example to manage the processing of all stake account (500000) that can take more than 1s and still allow processing incoming stake account from geyser during the leader schedule calculus, I split the account storage, one part contain all account to calculate the schedule and the other part cache the incoming one that are integrated after the processing. During the leader schedule process, the notifications are not needed. They are only needed for the next epoch.

modules/consensus/src/lib.rs
musitdev: Update done

PR #177, opened 2023-09-02
Quic proxy gm improvements

grooviegermanikus:
Changes based on feedback here https://github.com/blockworks-foundation/lite-rpc/pull/164#discussion_r1313079727

### DIFF
--- a/quic-forward-proxy/src/inbound/proxy_listener.rs
+++ b/quic-forward-proxy/src/inbound/proxy_listener.rs
@@ -6,7 +6,7 @@ use crate::tls_self_signed_pair_generator::SelfSignedTlsConfigProvider;
 use crate::util::FALLBACK_TIMEOUT;
 use anyhow::{anyhow, bail, Context};
 use log::{debug, error, info, trace, warn};
-use quinn::{Connection, Endpoint, ServerConfig, VarInt};
+use quinn::{Connecting, Endpoint, ServerConfig, VarInt};
 use solana_sdk::packet::PACKET_DATA_SIZE;
 use std::net::SocketAddr;
 use std::sync::Arc;
@@ -46,8 +46,7 @@ impl ProxyListener {
         while let Some(connecting) = endpoint.accept().await {
             let forwarder_channel_copy = forwarder_channel.clone();
             tokio::spawn(async move {
-                let connection = connecting.await.context("handshake").unwrap();
-                match Self::accept_client_connection(connection, forwarder_channel_copy).await {
+                match Self::accept_client_connection(connecting, forwarder_channel_copy).await {
                     Ok(()) => {
                         debug!("connection handles correctly");
                     }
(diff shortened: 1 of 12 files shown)


quic-forward-proxy/src/inbound/proxy_listener.rs
```diff
@@ -49,17 +44,9 @@ impl ProxyListener {
...
-                match Self::accept_client_connection(
-                    connection,
-                    forwarder_channel_copy,
-                    exit_signal,
-                )
-                .await
-                {
+                match Self::accept_client_connection(connecting, forwarder_channel_copy).await {
                     Ok(()) => {
                         debug!("connection handles correctly");
```
godmodegalactus, line 51: Is is normal that after accepting connection we do not do anything?

quic-forward-proxy/src/outbound/tx_forward.rs
```diff
@@ -218,16 +216,23 @@ pub async fn tx_forwarder(
...
     let mut to_shutdown = Vec::new();
     for (tpu_address, handle) in &*agents {
         if tpu_address == current_tpu_address {
             continue;
         }
 
-        let last_used_ms = handle.age_ms.load(Ordering::Relaxed);
+        let unused_period = {
+            let last_used_at = handle.last_used_at.read().await;
+            now - *last_used_at
```
godmodegalactus, line 232: should be saturating_sub instead/

godmodegalactus (changes requested):

quic-forward-proxy/src/inbound/proxy_listener.rs
grooviegermanikus: code is IMO okey, but naming is bad - will change _accept_ to _handle_

quic-forward-proxy/src/outbound/tx_forward.rs
grooviegermanikus: nono . that's Instant - Instant

godmodegalactus: Ahh yes.

quic-forward-proxy/src/util.rs
```diff
@@ -15,27 +13,3 @@ where
 {
     tokio::time::timeout(FALLBACK_TIMEOUT, future)
 }
-
```
godmodegalactus, line 18: this file should be removed right?

godmodegalactus (changes requested):

quic-forward-proxy/src/util.rs
grooviegermanikus: why that? should I inline the timeout?

godmodegalactus: Ok i guess its fine.

godmodegalactus (approved):

PR #194, opened 2023-09-13
using semaphore in tpu connection manager and removing unwanted code

godmodegalactus:


lite-rpc/src/lib.rs
```diff
@@ -28,7 +28,7 @@ pub const DEFAULT_FANOUT_SIZE: u64 = 10;
 #[from_env]
 pub const MAX_RETRIES: usize = 40;
 
-pub const DEFAULT_RETRY_TIMEOUT: u64 = 2;
+pub const DEFAULT_RETRY_TIMEOUT: u64 = 1;
```
grooviegermanikus, line 31: can this be _Duration_ typed or have at least a suffix and/or comment if it's seconds/milliseconds/etc.

services/src/tpu_utils/tpu_connection_manager.rs
```diff
@@ -156,13 +151,12 @@ impl ActiveConnection {
 
                     let task_counter = task_counter.clone();
                     let connection_pool = connection_pool.clone();
-
+                    let permit = task_counter.acquire_owned().await.expect("Should get permit");
```
grooviegermanikus, line 154: could you put a comment, what the expected behavior should be wrt backpressure? more transactions will be .recv'd and block the select-handler?
wouldn't it be better to keep the transactions in the channel?

lite-rpc/src/lib.rs
godmodegalactus: Duration is used in CLI as a default and Duration is incompatible with Clap

grooviegermanikus: k. got it, that's sad

grooviegermanikus (approved):
k

PR #195, opened 2023-09-13
restructuring core library into multiple directories

godmodegalactus:


core/src/traits/block_storage.rs
grooviegermanikus, line 1: module could be renamed to _block_storage_interface_ or so

core/src/stores/mod.rs
grooviegermanikus, line 1: you could add a comment to each _mod.rs_ file to explain the purpose/character of the contained mods
plus: explain the dependencies between then (e.g. no mod from _structures_ must ever use a type from _stores_)

core/src/stores/cluster_info.rs
grooviegermanikus, line 1: IMO all modules should have a consistent suffix: _cluster_info_store.rs_

core/src/types.rs
grooviegermanikus, line 1: this should IMO be removed and the respective type definitions should go to where it is used

godmodegalactus: I think this can remain as it is, because there are few types we defined here and used everywhere.

core/src/stores/mod.rs
godmodegalactus: Kind of explained what each mod has in one line. Do not want to start explaining the dependencies.

grooviegermanikus:
lgtm

musitdev (commented):
In my opinion, crate structure should be more organized by the function or domain that what it is here. You structure it by what it is; service, core function, ... So for example if you need core/storage you get all the dep of core/structure like leader that is not needed. This way to do can become problematic when you update the code. If for example I add a new function (vote account for example), I've to update every crate pass all test, be compatible with all dep and all developer of other function must update their code because I will update all crate: core, service, ...
I think we should start with a more domain oriented structure, like I propose in this PR: #173.
Each function is separated in its own crate, so all function code are independent.

godmodegalactus:
> In my opinion, crate structure should be more organized by the function or domain that what it is here. You structure it by what it is; service, core function, ... So for example if you need core/storage you get all the dep of core/structure like leader that is not needed. This way to do can become problematic when you update the code. If for example I add a new function (vote account for example), I've to update every crate pass all test, be compatible with all dep and all developer of other function must update their code because I will update all crate: core, service, ... I think we should start with a more domain oriented structure, like I propose in this PR: #173. Each function is separated in its own crate, so all function code are independent.

The core structures should be based on what the class is suppose to. I agree the services library should be based on what you have described in the docs. Could you start moving the services to reflect that please.

godmodegalactus:
If you want we can split services section into different libraries.

PR #196, opened 2023-09-14
improve the algorithm to schedule transaction replays

godmodegalactus:


services/src/transaction_replayer.rs
```diff
@@ -51,7 +57,14 @@ impl TransactionReplayer {
         tokio::spawn(async move {
             while let Some(mut tx_replay) = reciever.recv().await {
                 MESSAGES_IN_REPLAY_QUEUE.dec();
-                if Instant::now() < tx_replay.replay_at {
+                let now = Instant::now();
+                if now < tx_replay.replay_at {
+                    if tx_replay.replay_at > now + retry_after {
```
grooviegermanikus, line 62: nit: naming of _replay_after_ could imply both: replay after timestamp X or replay after elapsed time

PR #204, opened 2023-09-20
Fixing blockhash valid checks

godmodegalactus:
Creating method to check if blockhash is valid, checking it before sending the transaction


core/src/stores/data_cache.rs
```diff
@@ -43,6 +47,38 @@ impl DataCache {
...
+            .is_transaction_confirmed(&sent_transaction_info.signature)
+            || self
+                .block_store
+                .get_latest_block(CommitmentConfig::processed())
+                .await
+                .block_height
+                > sent_transaction_info.last_valid_block_height
+    }
+
+    pub fn new_for_tests() -> Self {
```
aniketfuryrocks, line 65: could add cfg test 🤔

godmodegalactus: means?

godmodegalactus: Cannot use cfg test, they are not in same crate.

https://stackoverflow.com/questions/72901643/cant-call-function-in-unit-test-from-integration-test-code

aniketfuryrocks (approved):

PR #206, opened 2023-09-21
add microbatching to proxy forward loop

grooviegermanikus:
add 40ms between next message in channel and draining the channel; limit the batch size to 50

this chart (metric _literpcproxy_batch_size_) shows the difference between no microbatching and 40ms: batch size of 1 is without microbatching:

Bildschirmfoto 2023-09-21 um 11 52 18


bench/transactions.csv
```diff
@@ -0,0 +1,901 @@
+signature,sent_slot,confirmed_slot,time_to_send_in_millis,time_to_confirm_in_millis
+2FLEisvJiqDrH8AA6Y2WKnPQdVGwuy31rQuyzfLd4diJ98kgHcPFbDdouMEWov3wtdZtJGX9sJWXNA9s5o1XuaNf,232084,232084,1,272
+2ouwbN9vAi1jQ2QSEsn3G1e5uWNdECXhB239xvbWTG6QnEMnQffDLMHyMpXap5CqCbK8hqb7fQ6mGUJnKCNN7eSJ,232084,232084,1,259
+5rSVfrKso2DSZxNFD3w16qR9aLdir3G1V4wuqcgozJwUp7iuivMzXAVSyPP6z3QZxiRhFTNLUBfL1b9x3xw3Fpev,232084,232084,2,251
```
godmodegalactus, line 4: Do not commit then bench file

quic-forward-proxy/src/outbound/tx_forward.rs
```diff
@@ -138,16 +144,26 @@ pub async fn tx_forwarder(
...
+                        // micro-batch
+                        // - after receiving one message wait a bit before draining the channel
+                        // - should be a fraction of slot time - i.e. small enough to not miss the slot
+                        // - check prometheus metric 'literpcproxy_batch_size'
+                        sleep(MICROBATCH_WINDOW).await;
+
                         let mut transactions_batch: Vec<Vec<u8>> = packet.transactions.clone();
 
-                        'more: while let Ok(more) = per_connection_receiver.try_recv() {
+                        'drain_loop: while let Ok(more) = per_connection_receiver.try_recv() {
```
godmodegalactus, line 155: why do you need to specify drain_loop, seems like goto which is not required right ?

godmodegalactus (changes requested):

bench/transactions.csv
grooviegermanikus: oops .sorry

quic-forward-proxy/src/outbound/tx_forward.rs
grooviegermanikus: it's not needed but ready more clearly IMO

aniketfuryrocks: imo this is a bad design. if u need labels to break out loops. could be done better

grooviegermanikus: no I do not need them.

compare it with this - the I scan that code I constantly look for the scope of breaks:
image

grooviegermanikus: I was staring a bit on the code, but so little room for improvement - can you suggest something?

godmodegalactus: I guess I read more clearly when i read continue and break logic.

godmodegalactus (approved):

grooviegermanikus:
Max advices to not do that because transactions are MTU sized

PR #208, opened 2023-09-21
proxy bench tools

grooviegermanikus:
added bench and test tools

* send_test_proxy_forward_requests: send transactions to proxy using the ProxyRequestFormat wire format
* bench_direct_quic: send transactions directly to TPU using QUIC
* rename main.rs to bench_rpc.rs - it uses lite RPC send_transaction method to send transactions


grooviegermanikus:
this will most likely conflict with  TPU client changes - will first focus on that and come back to this

PR #214, opened 2023-09-22
Multiplexing multiple endpoints

godmodegalactus:
Multiplexing grpc and rpc streams into one endpoint.


cluster-endpoints/src/grpc_subscription.rs
```diff
@@ -307,7 +307,7 @@ pub fn create_block_processing_task(
 }
 
 pub fn create_grpc_subscription(
```
grooviegermanikus, line 309: roc-based polling should really not go into a `grpc_subscription` method. maybe rename the method or use two

cluster-endpoints/src/multiplexing_subscriptions.rs
```diff
@@ -0,0 +1,112 @@
...
+    let (block_sx, blocks_notifier) = tokio::sync::broadcast::channel(10);
+    let mut endpoint_tasks = vec![];
+
+    let mut rpc_slot_notifier = rpc_endpoints.slot_notifier;
+    let mut grpc_slot_notifier = grpc_endpoint.slot_notifier;
+    let slot_multiplexer: AnyhowJoinHandle = tokio::spawn(async move {
+        let mut processed_slot = 0;
+        let mut estimated_slot = 0;
+        loop {
+            let (notification, is_grpc_update) = tokio::select! {
```
grooviegermanikus, line 32: pls to not use bool for `is_grpc_update` but enum instead 
(https://corrode.dev/blog/enums/)

cluster-endpoints/src/multiplexing_subscriptions.rs
```diff
@@ -0,0 +1,112 @@
...
+            if notification.processed_slot > processed_slot
+                || notification.estimated_processed_slot > estimated_slot
+            {
+                if is_grpc_update {
+                    GRPC_SLOT_UPDATE.set(1.0);
+                } else {
+                    GRPC_SLOT_UPDATE.set(0.0);
+                }
+                processed_slot = notification.processed_slot;
+                estimated_slot = notification.estimated_processed_slot;
```
grooviegermanikus, line 58: you never want to allow that the slots go back in time - right ? you could assert! that

`
assert!(notification.processed_slot >= processed_slot);
assert!(notification.estimated_processed_slot >= estimated_slot);
`

cluster-endpoints/src/grpc_subscription.rs
godmodegalactus: This is still under work, the rpc base polling system is just for vote-accounts and node-info, this will be replaced soon.

cluster-endpoints/src/multiplexing_subscriptions.rs
godmodegalactus: This makes sense if the boolean is in struct, but does not if it is a variable which is later used.

cluster-endpoints/src/multiplexing_subscriptions.rs
godmodegalactus: It is not i do not want to allow slots to go back in time, more like if rpc notification is late than grpc, so grpc has already progressed the slot, so we can ignore rpc notification.

PR #218, opened 2023-09-28
WIP: use gm tpuconnectionmanager for proxy

grooviegermanikus:


quic-forward-proxy/src/outbound/tpu_connection_manager.rs
```diff
@@ -0,0 +1,283 @@
...
+
+pub type WireTransaction = Vec<u8>;
+
+#[derive(Debug, Clone)]
+pub enum BroadcastMessage {
+    Transaction(WireTransaction),
+    Shutdown(Pubkey),
+}
+
+struct ActiveConnection {
```
grooviegermanikus, line 56: note: no `Clone` anymore which IMO is good

quic-forward-proxy/src/outbound/tpu_connection_manager.rs
```diff
@@ -0,0 +1,283 @@
...
+                            tokio::spawn(async move {
+                                // permit will be used to send all the transaction and then destroyed
+                                let _permit = permit;
+                                NB_QUIC_TASKS.inc();
+                                connection.send_transaction(tx_raw).await;
+                                NB_QUIC_TASKS.dec();
+                            });
+
+                        },
+                        Ok(BroadcastMessage::Shutdown(tpu_identity)) => {
```
grooviegermanikus, line 122: wdyt? sending _shutdown_ in line with transaction payload in same channel will remove race condition between transactions remaining in queue while connection thread is shutting down

quic-forward-proxy/src/outbound/tpu_connection_manager.rs
```diff
@@ -0,0 +1,283 @@
+#![allow(unused_imports)]
```
godmodegalactus, line 1: Can make changes to original code instead of coping it.

quic-forward-proxy/src/outbound/tpu_connection_manager.rs
```diff
@@ -0,0 +1,283 @@
...
+                self.endpoints.clone(),
+                *tpu_addr,
+                *tpu_identity,
+                connection_parameters,
+            );
+
+            let broadcast_receiver = self.broadcast_sender.subscribe();
+
+            active_connection.start_listening(broadcast_receiver, max_uni_stream_connections);
+            self.identity_to_active_connection.insert(
```
godmodegalactus, line 246: Connections are never removed from this map. They are always added.

quic-forward-proxy/src/outbound/tpu_connection_manager.rs
godmodegalactus: Ok i think, removes one extra channel.

quic-forward-proxy/src/outbound/tpu_connection_manager.rs
grooviegermanikus: eh! don't scare me ;) - look, I guess you are wrong:

> 2023-09-29T15:08:05.018721Z  INFO update_connections: solana_lite_rpc_quic_forward_proxy::outbound::tpu_connection_manager: identity_to_active_connection: 9 
...
2023-09-29T15:08:13.737999Z  INFO update_connections: solana_lite_rpc_quic_forward_proxy::outbound::tpu_connection_manager: identity_to_active_connection: 3

PR #221, opened 2023-10-03
Save all transactions in the cache, save transaction messages too, In…

godmodegalactus:
…crease transaction store size


cluster-endpoints/src/grpc_subscription.rs
```diff
@@ -192,6 +193,8 @@ fn process_block(
                 cu_requested,
                 prioritization_fees,
                 cu_consumed: compute_units_consumed,
+                blockhash: message.recent_blockhash().to_string(),
```
aniketfuryrocks, line 196: Latest block hash ?

aniketfuryrocks (approved):

PR #222, opened 2023-10-04
Epoch integration

musitdev:
Add epoch struct and calculus.
Add RPC get_epoch_info.


epoch/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+            Epoch {
+                epoch: 0,
+                slot_index: 1,
+                slots_in_epoch: 1000,
+                absolute_slot: 1,
+            },
+            slot1epoch
+        );
+
+        let slot1epoch = epoch_cache.get_epoch_at_slot(999);
```
godmodegalactus, line 109: variable name should be slot_epoch instead because it is not slot 1 epoch anymore.

epoch/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+use solana_rpc_client::nonblocking::rpc_client::RpcClient;
+use solana_sdk::epoch_info::EpochInfo;
+use solana_sdk::slot_history::Slot;
+use solana_sdk::sysvar::epoch_schedule::EpochSchedule;
+use std::sync::Arc;
+
+#[derive(Debug, Copy, Clone, PartialOrd, PartialEq, Eq, Ord)]
+pub struct Epoch {
+    pub epoch: u64,
+    pub slot_index: u64,
```
godmodegalactus, line 12: I do not see why we need to create a new library for epoch, this file can go in core library.

lite-rpc/src/bridge.rs
```diff
@@ -334,4 +335,20 @@ impl LiteRpcServer for LiteBridge {
             Ok(None)
         }
     }
+
+    async fn get_epoch_info(&self) -> crate::rpc::Result<EpochInfo> {
```
godmodegalactus, line 339: Does not look like the same definition in existing RPC. Should be somewhat similar, I see that there is missing commitment and minContext Slot.

https://docs.solana.com/api/http#getepochinfo

bench/src/cli.rs
```diff
@@ -20,4 +20,7 @@ pub struct Args {
     pub lite_rpc_addr: String,
     #[arg(short = 't', long, default_value_t = String::from("transactions.csv"))]
     pub transaction_save_file: String,
+    // choose between small (179 bytes) and large (1186 bytes) transactions
+    #[arg(short = 'L', long, default_value_t = false)]
```
godmodegalactus, line 24: bad rebase.

epoch/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    pub async fn bootstrap_epoch(rpc_client: &RpcClient) -> Result<EpochCache, ClientError> {
+        let res_epoch = rpc_client
+            .get_account(&solana_sdk::sysvar::epoch_schedule::id())
+            .await?;
+        let Some(SysvarAccountType::EpochSchedule(epoch_schedule)) = bincode::deserialize(&res_epoch.data[..])
+        .ok()
+        .map(SysvarAccountType::EpochSchedule) else {
+        panic!("SysvarAccountType::EpochSchedule can(t be deserilized. Eoch can't be calculated.");
```
godmodegalactus, line 66: avoid panic and return Err.

godmodegalactus (changes requested):

epoch/src/lib.rs
musitdev: I didn't know where to put it so as it's independent, I put in a crate until we decide where to put.
I put it in core/structures?

lite-rpc/src/bridge.rs
musitdev: I take the solana_rpc_client signature: https://docs.rs/solana-client/latest/solana_client/nonblocking/rpc_client/struct.RpcClient.html#method.get_epoch_info
I see that there's a [get_epoch_info_with_commitment](https://docs.rs/solana-client/latest/solana_client/nonblocking/rpc_client/struct.RpcClient.html#method.get_epoch_info_with_commitment).
I use the solana client API or the RPC doc?

bench/src/cli.rs
musitdev: Yes there's some issue with the rebase. I try to restart, but I think I've to recreate a new PR.

PR #223, opened 2023-10-04
Epoch integration

musitdev:
Add epoch struct and calculus.
Add RPC get_epoch_info.

I add the Epoch info test. I change the default RPC port that wasn't 8899.


godmodegalactus (approved):

lite-rpc/src/bridge.rs
```diff
@@ -334,4 +335,23 @@ impl LiteRpcServer for LiteBridge {
...
     }
+
+    async fn get_epoch_info(
+        &self,
+        _config: Option<RpcContextConfig>,
+    ) -> crate::rpc::Result<EpochInfo> {
+        let block_info = self
+            .data_cache
+            .block_information_store
+            .get_latest_block_info(CommitmentConfig::finalized())
```
godmodegalactus, line 346: Here commitment should come from the config

godmodegalactus (changes requested):

lite-rpc/src/bridge.rs
musitdev: Ah yes, you're right and it's not the right commitment.

tests/client.test.ts
```diff
@@ -45,3 +45,10 @@ test('send and confirm transaction', async () => {
     await sendAndConfirmTransaction(connection, tx, [payer]);
 });
 
+
+test('get epoch info', async () => {
+    const {epoch, absoluteSlot, slotIndex, slotsInEpoch} = await connection.getEpochInfo();
```
godmodegalactus, line 50: Could also try adding tests with different commitments.

godmodegalactus (approved):

PR #227, opened 2023-10-05
Saving block and transactions in postgres table

godmodegalactus:


history/src/block_stores/postgres_block_store.rs
```diff
@@ -1 +1,106 @@
...
+        // create schema for new epoch
+        let session = self
+            .session_cache
+            .get_session()
+            .await
+            .expect("should get new postgres session");
+
+        let statement = format!("CREATE SCHEMA {};", schema);
+        if let Err(e) = session.execute(&statement, &[]).await {
+            log::error!("Error creating a schema for {} error {}", schema, e);
```
musitdev, line 42: I think the errors should be return because is the db is not created there's no need to store block and other actions must be done ASAP.

history/src/block_stores/postgres_block_store.rs
```diff
@@ -1 +1,106 @@
...
+            .transactions
+            .iter()
+            .map(|x| PostgresTransaction::new(x, slot))
+            .collect_vec();
+        let postgres_block = PostgresBlock::from(&block);
+
+        let epoch = self.epoch_cache.get_epoch_at_slot(slot);
+        let schema = format!("EPOCH_{}", epoch.epoch);
+        if current_epoch == 0 || current_epoch < epoch.epoch {
+            self.postgres_data.write().await.current_epoch = epoch.epoch;
```
musitdev, line 77: You update the epoch and not the from and to slot range.
Perhaps it's easier to use the start and end slot function of epoch to have the range and not using a specific attribute from_slot and to_slot?

history/src/postgres/postgres_block.rs
```diff
@@ -1,45 +1,85 @@
...
+        args.push(&self.blockhash);
+        args.push(&self.block_height);
+        args.push(&self.parent_slot);
+        args.push(&self.block_time);
+        args.push(&self.previous_blockhash);
+        args.push(&self.rewards);
+
+        PostgresSession::multiline_query(&mut query, NB_ARUMENTS, 1, &[]);
+        if let Err(e) = postgres_session.execute(&query, &args).await {
+            log::error!(
```
musitdev, line 77: The same perhaps it's better not to hide the error and return it.

history/src/postgres/postgres_transaction.rs
```diff
@@ -32,6 +33,85 @@ impl From<&TransactionInfo> for PostgresTransaction {
...
+            r#"
+                INSERT INTO {}.TRANSACTIONS 
+                (signature, slot, err, cu_requested, prioritization_fees, cu_consumed, recent_blockhash, message)
+                VALUES
+            "#,
+            schema
+        );
+
+        PostgresSession::multiline_query(&mut query, NB_ARUMENTS, transactions.len(), &[]);
+        if let Err(e) = postgres_session.execute(&query, &args).await {
```
musitdev, line 100: The same, return the error?

history/src/block_stores/postgres_block_store.rs
godmodegalactus: from_slot and to_slot stores which slots are stored in the block store. It can change for in memory blocksotre where we store only last few blocks. Similarly it will change for faithful as the faithful will take some time to store the latest epoch. The epoch thing is particular to postgres because we will create a new schema per epoch. And here we get current epoch from the block slot.

history/src/postgres/postgres_transaction.rs
godmodegalactus: I will do the changes.

musitdev (approved):

history/src/postgres/postgres_block.rs
```diff
@@ -1,45 +1,83 @@
...
-    pub commitment_config: i8,
     pub previous_blockhash: String,
     pub rewards: Option<String>,
 }
 
-impl SchemaSize for PostgresBlock {
-    const DEFAULT_SIZE: usize = 4 * 8;
-    const MAX_SIZE: usize = Self::DEFAULT_SIZE + 8;
-}
+const NB_ARUMENTS: usize = 7;
```
grooviegermanikus, line 17: typo?

history/src/postgres/postgres_block.rs
```diff
@@ -1,45 +1,83 @@
...
-            commitment_config: commitment as i8,
-            previous_blockhash: value.previous_blockhash,
+            previous_blockhash: value.previous_blockhash.clone(),
             rewards,
         }
     }
 }
+
+impl PostgresBlock {
+    pub fn create_statement(schema: &String) -> String {
```
grooviegermanikus, line 40: you could add a defensive check for the schema format. e.g.
`assert!(schema.starts_with("lite_rpc_foo"))`

history/src/postgres/postgres_block.rs
```diff
@@ -1,45 +1,83 @@
...
+        format!(
+            "
+            CREATE TABLE {}.BLOCKS (
```
grooviegermanikus, line 43: pls clarify the commitment level; maybe by naming the primary key `slot_confirmed` or adding a comment somewhere or by choosing a sound schema name like `lite_rpc_blocks_confirmed`

history/src/block_stores/postgres_block_store.rs
```diff
@@ -1 +1,101 @@
...
+}
+
+pub struct PostgresBlockStore {
+    session_cache: PostgresSessionCache,
+    epoch_cache: EpochCache,
+    postgres_data: Arc<RwLock<PostgresData>>,
+}
+
+impl PostgresBlockStore {
+    pub async fn start_new_epoch(&self, schema: &String) -> Result<()> {
```
grooviegermanikus, line 33: IMO the schema should not be passed here as a parameter but constructed inside the method

history/src/postgres/postgres_block.rs
godmodegalactus: We only save finalized blocks in postgres

godmodegalactus:
@grooviegermanikus I will address comments in next PR.

core/src/traits/block_storage_interface.rs
```diff
@@ -7,11 +8,12 @@ use std::{ops::Range, sync::Arc};
 #[async_trait]
 pub trait BlockStorageInterface: Send + Sync {
     // will save a block
-    async fn save(&self, block: ProducedBlock);
+    async fn save(&self, block: ProducedBlock) -> Result<()>;
```
grooviegermanikus, line 11:
```suggestion
    async fn save(&self, block: &ProducedBlock) -> Result<()>;
```

history/src/postgres/postgres_block.rs
```diff
@@ -1,45 +1,83 @@
...
+                slot BIGINT PRIMARY KEY,
+                blockhash STRING NOT NULL,
```
grooviegermanikus, line 45: what is this _string_ type?

history/src/postgres/postgres_transaction.rs
```diff
@@ -32,6 +32,81 @@ impl From<&TransactionInfo> for PostgresTransaction {
...
+            signature CHAR(88) NOT NULL,
+            slot BIGINT, 
+            err STRING,
+            cu_requested BIGINT,
+            prioritization_fees BIGINT,
+            cu_consumed BIGINT,
+            recent_blockhash STRING NOT NULL,
+            message STRING NOT NULL,
+            PRIMARY KEY (signature)
+            CONSTRAINT fk_transactions FOREIGN KEY (slot) REFERENCES {}.BLOCKS(slot);
```
grooviegermanikus, line 52: the trailing semicolon is wrong I think

grooviegermanikus, line 52: shouldn't the foreign key name contain the column name `slot` ?

core/src/traits/block_storage_interface.rs
```diff
@@ -7,11 +8,12 @@ use std::{ops::Range, sync::Arc};
...
     // will get a block
-    async fn get(&self, slot: Slot, config: RpcBlockConfig) -> Option<ProducedBlock>;
+    async fn get(&self, slot: Slot, config: RpcBlockConfig) -> Result<ProducedBlock>;
     // will get range of slots that are stored in the storage
     async fn get_slot_range(&self) -> Range<Slot>;
```
grooviegermanikus, line 15: range means a consecutive sequence of slots

history/src/block_stores/multiple_strategy_block_store.rs
```diff
@@ -60,46 +61,47 @@ impl MultipleStrategyBlockStorage {
 
 #[async_trait]
 impl BlockStorageInterface for MultipleStrategyBlockStorage {
-    async fn save(&self, block: ProducedBlock) {
+    async fn save(&self, block: ProducedBlock) -> Result<()> {
         let slot = block.slot;
         let commitment = Commitment::from(block.commitment_config);
         match commitment {
             Commitment::Confirmed | Commitment::Processed => {
```
grooviegermanikus, line 68: nit: pls sort by logical order

history/src/block_stores/multiple_strategy_block_store.rs
```diff
@@ -60,46 +61,47 @@ impl MultipleStrategyBlockStorage {
...
                         }
                     }
-                    None => self.inmemory_for_storage.save(block.clone()).await,
+                    Err(_) => self.inmemory_for_storage.save(block.clone()).await?,
                 }
-                self.persistent_block_storage.save(block).await;
+                self.persistent_block_storage.save(block).await?;
             }
         };
         if slot > self.last_confirmed_slot.load(Ordering::Relaxed) {
```
grooviegermanikus, line 86: is that correct? what if commitment==Processed?

history/src/block_stores/multiple_strategy_block_store.rs
```diff
@@ -60,46 +61,47 @@ impl MultipleStrategyBlockStorage {
 
 #[async_trait]
 impl BlockStorageInterface for MultipleStrategyBlockStorage {
-    async fn save(&self, block: ProducedBlock) {
+    async fn save(&self, block: ProducedBlock) -> Result<()> {
         let slot = block.slot;
         let commitment = Commitment::from(block.commitment_config);
```
grooviegermanikus, line 66: why is that property called _commitment_config_ ? should be the commitment status/level?

history/src/block_stores/inmemory_block_store.rs
```diff
@@ -52,12 +53,18 @@ impl InmemoryBlockStore {
 
 #[async_trait]
 impl BlockStorageInterface for InmemoryBlockStore {
-    async fn save(&self, block: ProducedBlock) {
+    async fn save(&self, block: ProducedBlock) -> anyhow::Result<()> {
```
grooviegermanikus, line 56: we should prevent a slot going back in the states model

again: IMO the modeling is not reflecting the state model of slot->commitment_status correctly

history/src/postgres/postgres_block.rs
```diff
@@ -1,45 +1,83 @@
...
+    }
+
+    pub async fn save(
+        &self,
+        postgres_session: &PostgresSession,
+        schema: &String,
+    ) -> anyhow::Result<()> {
+        let mut query = format!(
+            r#"
+            INSERT INTO {}.BLOCKS (slot, blockhash, block_height, parent_slot, block_time, previous_blockhash, rewards) VALUES
```
grooviegermanikus, line 65: this can be called only once? from the call-chain I'd assume that `.save` gets called multiple times

history/src/postgres/postgres_block.rs
```diff
@@ -1,45 +1,83 @@
...
 
-impl From<ProducedBlock> for PostgresBlock {
-    fn from(value: ProducedBlock) -> Self {
-        let commitment = Commitment::from(&value.commitment_config);
+impl From<&ProducedBlock> for PostgresBlock {
+    fn from(value: &ProducedBlock) -> Self {
         let rewards = value
             .rewards
             .as_ref()
             .map(|x| BASE64.serialize(x).ok())
```
grooviegermanikus, line 24: maybe we should add a version tag: I use this pattern: `v1#ZDM0ODcyNDUwMjViMjM0NGJhNDc4ZDgyYTVjODM5OGQ4Nzg1OWQ3N2IzZmE3NDAyMmNlYTAyYzljYjMwYjA1OAo=`

history/src/postgres/postgres_transaction.rs
```diff
@@ -32,6 +32,81 @@ impl From<&TransactionInfo> for PostgresTransaction {
...
             recent_blockhash: value.recent_blockhash.clone(),
             message: value.message.clone(),
+            slot: slot as i64,
+        }
+    }
+
+    pub fn create_statement(schema: &String) -> String {
+        format!(
+            "\
+        CREATE TABLE {}.TRANSACTIONS (
```
grooviegermanikus, line 42: possible performance improvements:

### Cluster index (recommended)
tldr: will physically collocate data based on values (.slot)

> CREATE INDEX idx_tx_slot ON TRANSACTIONS(slot);
CLUSTER TRANSACTIONS using idx_tx_slot
VACUUM FULL TRANSACTIONS
-- https://www.cybertec-postgresql.com/en/cluster-improving-postgresql-performance/


### GIN/Bitmap
will use bitmap index instead of b-tree which is more optimal for large buckets of data (like transactions per slots)

> INDEX idx_tx_slot  ON TRANSACTIONS USING gin (slot);
-- http://hlinnaka.iki.fi/2014/03/28/gin-as-a-substitute-for-bitmap-indexes/

history/src/block_stores/postgres_block_store.rs
```diff
@@ -1 +1,101 @@
...
+            .iter()
+            .map(|x| PostgresTransaction::new(x, slot))
+            .collect_vec();
+        let postgres_block = PostgresBlock::from(&block);
+
+        let epoch = self.epoch_cache.get_epoch_at_slot(slot);
+        let schema = format!("EPOCH_{}", epoch.epoch);
+        if current_epoch == 0 || current_epoch < epoch.epoch {
+            self.postgres_data.write().await.current_epoch = epoch.epoch;
+            self.start_new_epoch(&schema).await?;
```
grooviegermanikus, line 72: could result in race condition,
to work around it, you could write

> CREATE SCHEMA IF NOT EXISTS ...

history/src/block_stores/postgres_block_store.rs
```diff
@@ -1 +1,101 @@
...
+use solana_sdk::{slot_history::Slot, stake_history::Epoch};
+use tokio::sync::RwLock;
+
+use crate::postgres::{
+    postgres_block::PostgresBlock, postgres_session::PostgresSessionCache,
+    postgres_transaction::PostgresTransaction,
+};
+
+#[derive(Default, Clone, Copy)]
+pub struct PostgresData {
```
grooviegermanikus, line 20: can't we somehow remove this extra state?

history/src/block_stores/postgres_block_store.rs
```diff
@@ -1 +1,101 @@
...
+    from_slot: Slot,
+    to_slot: Slot,
+    current_epoch: Epoch,
+}
+
+pub struct PostgresBlockStore {
+    session_cache: PostgresSessionCache,
```
grooviegermanikus, line 27: did you consider using an existing pool like [Deadpool](https://crates.io/crates/deadpool-postgres)

history/src/postgres/postgres_transaction.rs
```diff
@@ -32,6 +32,81 @@ impl From<&TransactionInfo> for PostgresTransaction {
...
+          );
+        ",
+            schema, schema
+        )
+    }
+
+    pub async fn save_transactions(
+        postgres_session: &PostgresSession,
+        schema: &String,
+        transactions: &[Self],
```
grooviegermanikus, line 62: how many transactions to are we expecting here? could you document that? are the figures reasonably low to fit in one statement?

history/src/postgres/postgres_transaction.rs
```diff
@@ -32,6 +32,81 @@ impl From<&TransactionInfo> for PostgresTransaction {
...
             message: value.message.clone(),
+            slot: slot as i64,
+        }
+    }
+
+    pub fn create_statement(schema: &String) -> String {
+        format!(
+            "\
+        CREATE TABLE {}.TRANSACTIONS (
+            signature CHAR(88) NOT NULL,
```
grooviegermanikus, line 43: _CHAR_ is not recommended:

> Tip
There is no performance difference among these three types, apart from increased storage space when using the blank-padded type, and a few extra CPU cycles to check the length when storing into a length-constrained column. While character(n) has performance advantages in some other database systems, there is no such advantage in PostgreSQL; in fact character(n) is usually the slowest of the three because of its additional storage costs. In most situations text or character varying should be used instead.

https://www.postgresql.org/docs/current/datatype-character.html

history/src/postgres/postgres_block.rs
```diff
@@ -1,45 +1,83 @@
...
         }
     }
 }
+
+impl PostgresBlock {
+    pub fn create_statement(schema: &String) -> String {
+        format!(
+            "
+            CREATE TABLE {}.BLOCKS (
+                slot BIGINT PRIMARY KEY,
```
grooviegermanikus, line 44: would suggest to consistently declare the _PRIMARY KEY_ later:

> slot BIGINT NOT NULL,
...
PRIMARY KEY (slot)

history/src/postgres/postgres_block.rs
```diff
@@ -1,45 +1,83 @@
...
+                blockhash STRING NOT NULL,
+                leader_id STRING,
```
grooviegermanikus, line 46: guess it's too late but I in my models I use the convention that `foo_id` refers to an surrogate id and not to something outside the system boundaries
for the outside id's I use `foo_ident` or the naming convention there (if it's not id ofc ;) )

history/src/postgres/postgres_block.rs
grooviegermanikus, line 43: maybe these DDL strings could be include from an external file

Issue #199, opened 2023-09-15
Communication problems (GSO/IO) between lite-rpc and quic-forward-proxy on IPv6 private network


riordanp:
When deploying lite-rpc and quic-forward-proxy to Fly on separate VMs, we noticed submitted TXs timing out.
After some debugging, it seems [this call](https://github.com/blockworks-foundation/lite-rpc/blob/main/services/src/tpu_utils/quinn_auto_reconnect.rs#L48) to write_all on the SendStream can produce this error:

```
sendmsg error: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }, Transmit: { destination: [fdaa:0:a53d:a7b:17b:3cd9:fe0f:2]:11111, src_ip: None, enc: Some(Ect0), len: 1291, segment_size: Some(1200) }
```

After which the future never returns, blocking any further messages from the broadcast_receiver being processed. After restarting lite-rpc maybe one or two batches are successfully forwarded before this error occurs and no more batches are forwarded.

Adding a [timeout](https://github.com/blockworks-foundation/lite-rpc/commit/433550d004b0373d0bcaa39d86fe15203b360585) to this future allows sending to continue, but produces the following error logs in quic-forward-proxy:
```
ERROR handle_client_connection: solana_lite_rpc_quic_forward_proxy::inbound::proxy_listener: failed to accept stream: timed out
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Read(ConnectionLost(TimedOut))', quic-forward-proxy/src/inbound/proxy_listener.rs:106:85
ERROR solana_lite_rpc_quic_forward_proxy::inbound::proxy_listener: failed handling connection from client: error accepting stream - skip
```

Tests on IPv4 and local IPv6 outside of the Fly network have been successful, pointing to an incompatibility with the Fly IPv6 private network, or kernel network parameters on the VMs.

grooviegermanikus:
_sendmsg_ is implemented in quinn-udp to allow for large package sizes

we suspect that the issue is related to _Generic Segmentation Offload_ available in recent OS's

grooviegermanikus:
Quinn added opt-out to to 0.9 https://github.com/quinn-rs/quinn/pull/1671

grooviegermanikus:
opened [PR](https://github.com/blockworks-foundation/lite-rpc/pull/228) to allow to completely disable GSO on Quinn-Level

PR #228, opened 2023-10-09
disable gso (generic segmentation offload)

grooviegermanikus:
workaround for [broken GSO detection on fly.io](https://github.com/blockworks-foundation/lite-rpc/issues/199) - will disable GSO completely; quinn will not try to detect GSO support on the OS

**default** is to enable GSO (autodetected)

note: requires quinn-proto 0.9.6

Activate this feature by setting the env variable:
> DISABLE_GSO=true bin/lite-rpc ...

on deprecation:
the opt-out for GSO is only required as long as GSO detection is broken (currently at fly.io and quinn 0.9 [see](https://github.com/blockworks-foundation/lite-rpc/issues/199) ) - would recommend to remove this flag as soon at there is a fix


riordanp (approved):
LGTM. Updated env on fly, setting the new flag ready for this to merge

PR #229, opened 2023-10-10
Stake and vote management. Integration of the main loop in LiteRPC

musitdev:
First PR of stake and vote management.
 * Main loop that process geyser account notification.
 * Entry point for stake, vote and leader schedule calculus.
 * Integration in the current LiteRpc architecture.
    - add leader schedule to share data
    - leader_shedule RPC
    - start process notification loop.
    - all leader schedule modules definition
   
This PR validate the integration of the vote and stake management in current Lite RPC version.

PR #230, opened 2023-10-12
Stakes and votes management: add leader schedule calculus

musitdev:
This PR add the leader calculus to the #229 one.
The leader calculus is done at the end of the epoch on the stake and vote account stored and updated during the epoch.
It is done in a separate task, so I use that specific container call TakableMap that allow to take the content but allow caching update that arrive during the calculus and apply then when the content return (merge).
This process: Take, calculate in a task, merge is done using a simple state machine using LeaderScheduleEvent.


musitdev:
All unused are correct in the next pull request.

PR #232, opened 2023-10-18
Stakes and votes management: add stake and schedule bootstrap

musitdev:
Add stake and vote account bootstrap from the RPC getProgramAccount.
Add leader schedule, save at epoch change and reload at start if present.

### DIFF
--- a/stake_vote/src/lib.rs
+++ b/stake_vote/src/lib.rs
@@ -1,4 +1,5 @@
 use crate::account::AccountPretty;
+use crate::bootstrap::BootstrapEvent;
 use crate::leader_schedule::LeaderScheduleGeneratedData;
 use futures::Stream;
 use futures_util::stream::FuturesUnordered;
@@ -36,7 +37,9 @@ pub async fn start_stakes_and_votes_loop(
     rpc_client: Arc<RpcClient>,
     grpc_url: String,
 ) -> anyhow::Result<tokio::task::JoinHandle<()>> {
+    log::info!("Start Stake and Vote loop.");
     let mut account_gyzer_stream = subscribe_geyzer(grpc_url).await?;
+    log::info!("Stake and Vote geyzer subscription done.");
     let jh = tokio::spawn(async move {
         //Stake account management struct
         let mut stakestore = stake::StakeStore::new(STAKESTORE_INITIAL_CAPACITY);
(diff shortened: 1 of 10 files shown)


stake_vote/src/lib.rs
```diff
@@ -36,7 +37,9 @@ pub async fn start_stakes_and_votes_loop(
     rpc_client: Arc<RpcClient>,
     grpc_url: String,
 ) -> anyhow::Result<tokio::task::JoinHandle<()>> {
```
aniketfuryrocks, line 39: Use AnyhowJoinHandle

stake_vote/src/lib.rs
```diff
@@ -36,7 +37,9 @@ pub async fn start_stakes_and_votes_loop(
...
+    log::info!("Start Stake and Vote loop.");
     let mut account_gyzer_stream = subscribe_geyzer(grpc_url).await?;
+    log::info!("Stake and Vote geyzer subscription done.");
     let jh = tokio::spawn(async move {
```
aniketfuryrocks, line 43: Just return here, or preferably make the function async and let the caller decide whether to spawn a new task or not. I understand there is a loop here

stake_vote/src/utils.rs
```diff
@@ -21,6 +27,52 @@ pub async fn get_current_epoch(data_cache: &DataCache) -> LiteRpcEpoch {
...
+    };
+    let serialized_stakes = serde_json::to_string(&save_stakes).unwrap();
+    let mut file = File::create(base_file_path).unwrap();
+    file.write_all(serialized_stakes.as_bytes()).unwrap();
+    file.flush().unwrap();
+    Ok(())
+}
+
 //Takable struct code
 pub trait TakableContent<T>: Default {
```
aniketfuryrocks, line 77: This seems to be de coupled from this crate. Move this structure to core/src/structures.

stake_vote/src/lib.rs
musitdev: I understand the idea that we can decide how to start a task. I think here it's a whole. There's some init and then the loop task must be started. If not, there's no need to call the method. 
The other possibility is to use a struct that has an init method and a start loop method. It adds complexity and add the possibility to start the loop without doing the init. 
I think it's the same as RPC or Geyser subscription (https://github.com/blockworks-foundation/lite-rpc/blob/fa0433f74a3e0b89d5d93dd5fc4892e08dc0d1e9/cluster-endpoints/src/grpc_subscription.rs#L272). It's part the implementation and can't be removed.

stake_vote/src/utils.rs
musitdev: It's a specific pattern implementation use only by vote and stake. I prefer to let it there and move it if it's needed in another crate. In fact, it's going to have a lot of change and to be very specific to the stake and vote data management.

stake_vote/src/lib.rs
musitdev: OK, I'll do the change in the next PR because I've done a lot of change since and it's easier to do it in the current version of the code.

PR #243, opened 2023-11-13
Update grpc version to 0.11.0 so that we can use the latest triton se…

godmodegalactus:
…rvice

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,5 +60,5 @@ solana-lite-rpc-cluster-endpoints = {path = "cluster-endpoints", version="0.2.3"
 solana-lite-rpc-history = {path = "history", version="0.2.3"}
 
 async-trait = "0.1.68"
-yellowstone-grpc-client = "1.9.0"
-yellowstone-grpc-proto = "1.9.0"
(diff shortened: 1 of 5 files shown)


musitdev (approved):

Cargo.toml
```diff
@@ -60,5 +60,5 @@ solana-lite-rpc-cluster-endpoints = {path = "cluster-endpoints", version="0.2.3"
 solana-lite-rpc-history = {path = "history", version="0.2.3"}
 
 async-trait = "0.1.68"
-yellowstone-grpc-client = "1.9.0"
-yellowstone-grpc-proto = "1.9.0"
+yellowstone-grpc-client = "1.11.0"
```
grooviegermanikus, line 63: I wonder how the project's git tagging system works?
how do we know if we get`v1.11.0+solana.1.17.4` vs `v1.11.0+solana.1.16.17`?

godmodegalactus: Yes it is quite difficult to know with grpc plugin i agree. I think it depends on solana grpc plugin. If it was compiled with solana 1.17.4 or with v1.11.0 because anyways we want to get what is the version of grpc on the solana validator

PR #246, opened 2023-11-23
Cli + Config

aniketfuryrocks:


lite-rpc/Cargo.toml
```diff
@@ -16,7 +16,7 @@ solana-rpc-client-api = { workspace = true }
 solana-transaction-status = { workspace = true }
 solana-version = { workspace = true }
 serde = { workspace = true }
-serde_json = { workspace = true }
+serde_json = "1.0.108"
```
godmodegalactus, line 19: why not change the version in workspace itself.

config_parser/src/lib.rs
```diff
@@ -0,0 +1,217 @@
+use proc_macro::TokenStream;
+use quote::{quote, ToTokens};
+use syn::{Data, Expr};
+
+struct Arg<'a> {
```
godmodegalactus, line 5: Why do you have to do this, why cant you do the same as yellowstone-grpc justing using serde parse. It would be simpler to maintain rather than doing some custom stuff ?

https://github.com/rpcpool/yellowstone-grpc/blob/master/yellowstone-grpc-geyser/src/config.rs

config_parser/src/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+    name: &'a syn::Ident,
+    ty: &'a syn::Type,
+    desc: Option<Expr>,
+    default: Option<Expr>,
+    short: Option<Expr>,
```
godmodegalactus, line 10: short is no longer required, the config file should be readable.

godmodegalactus (changes requested):

config_parser/src/lib.rs
godmodegalactus: This does not even enable to create sections, that we had discussed about.

grooviegermanikus (changes requested):
@aniketfuryrocks  could you provide+commit an example for local config.json please

PR #247, opened 2023-11-23
simple config, env and cli

aniketfuryrocks:


lite-rpc/src/cli.rs
```diff
@@ -1,42 +1,134 @@
...
+        let args = Args::parse();
+
+        let config_path = if let Some(config) = &args.config {
+            config
+        } else {
+            let default_config_path = "config.json";
+
+            // check if config.json exists in current directory
+            if tokio::fs::metadata(default_config_path).await.is_err() {
+                return Ok(serde_json::from_str("{}").unwrap());
```
godmodegalactus, line 68: Should return error, config file should exist.

lite-rpc/src/cli.rs
```diff
@@ -1,42 +1,134 @@
...
+            .context("Error reading config file")?;
+
+        let mut config: Config =
+            serde_json::from_str(&config).context("Error parsing config file")?;
+
+        if args.identity_keypair.is_some() {
+            config.identity_keypair = args.identity_keypair;
+        }
+
+        config.enable_postgres = env::var("PG_ENABLED").is_ok();
```
godmodegalactus, line 90: Move PG related config to the configuration file.

godmodegalactus (changes requested):
Looks good.

lite-rpc/src/cli.rs
aniketfuryrocks: There are defaults populated so in case there is no config passed in the cli and no default config 'config.json' then use the defaults

godmodegalactus:
squash all commits into one, and then we can approve.

README.md
```diff
@@ -58,17 +58,24 @@ Find a new file named `metrics.csv` in the project root.
...
-| `IDENTITY`        | Staked validator identity keypair              | no                  |
-| `PG_ENABLED`      | Set to anything but 'false' to enable Postgres | no                  |
-| `PG_CONFIG`       | Postgres Connection Config                     | if postgres enabled |
-| `CA_PEM_B64`      | Base64 encoded `ca.pem`                        | if postgres enabled |
-| `CLIENT_PKS_B64`  | Base64 encoded `client.pks`                    | if postgres enabled |
-| `CLIENT_PKS_PASS` | Password to `client.pks`                       | if postgres enabled |
-| `DISABLE_GSO`     | Disable GSO completely                         | no                  |
+Thank you for providing the default values. Here's the updated table with the default values for the environment variables based on the additional information:
+
+| Environment Variable     | Purpose                                                          | Required?           | Default Value                                  |
```
godmodegalactus, line 63: Should provide a config file example instead of environment variables.

godmodegalactus (approved):

PR #252, opened 2023-12-04
-D clippy

aniketfuryrocks:


musitdev (approved):
I'm not good a shell and git CI, but as It works for me, it's ok.

PR #255, opened 2023-12-19
grpc multiplexing integrate

grooviegermanikus:
Use auto-reconnect and multiplex utilities from https://github.com/blockworks-foundation/geyser-grpc-connector to allow a second gRPC connection.

### Configuration
Up to 4 connections can be configured from **config.json** with overrides from **ENV**:
```
"use_grpc": true,
  "grpc_addr": "http://any.rpcpool.com",
  "grpc_x_token": "121212121212",
  "grpc_addr2": "http://some.rpcpool.com",
  "grpc_x_token2": null,
```
env variables:
> ....
> GRPC_ADDR2=
> GRPC_X_TOKEN2=mytoken


lite-rpc/src/main.rs
```diff
@@ -295,3 +308,11 @@ fn parse_host_port(host_port: &str) -> Result<SocketAddr, String> {
         Ok(addrs[0])
     }
 }
+
+// http://mango.rpcpool.com/c232ab232ba2323
+fn obfuscate_rpcurl(rpc_addr: &str) -> String {
```
grooviegermanikus, line 313: do we need this

lite-rpc/src/cli.rs
```diff
@@ -46,6 +46,10 @@ pub struct Config {
     pub grpc_addr: String,
     #[serde(default)]
     pub grpc_x_token: Option<String>,
+    #[serde(default)]
```
godmodegalactus, line 49: Make it Vec<String> instead of adding another argument.
Example : 
```
#[arg(short, long, value_delimiter = ',')]
 pub grpc_addr: Vec<String>,

```

godmodegalactus (approved):

godmodegalactus (changes requested):
Changes as requested

lite-rpc/src/cli.rs
grooviegermanikus: reworked it, but not as vector .. that's really an headache

godmodegalactus (approved):

PR #267, opened 2024-01-08
Reduce network by only fetching block meta instead of whole block

grooviegermanikus:
changes:
* renamed stuff
* added cleanup for map
* added logging
* remove obsolete match for Block

### DIFF
--- a/cluster-endpoints/src/grpc_multiplex.rs
+++ b/cluster-endpoints/src/grpc_multiplex.rs
@@ -104,12 +104,11 @@ pub fn create_grpc_multiplex_blocks_subscription(
 
             let sender = producedblock_sender;
             let mut cleanup_tick = tokio::time::interval(Duration::from_secs(5));
-            let mut current_slot: Slot = 0;
+            let mut last_finalized_slot: Slot = 0;
             loop {
                 tokio::select! {
                     confirmed_block = confirmed_blocks_stream.next() => {
                         let confirmed_block = confirmed_block.expect("confirmed block from stream");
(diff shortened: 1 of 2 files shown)


cluster-endpoints/src/grpc_multiplex.rs
```diff
@@ -68,28 +77,75 @@ pub fn create_grpc_multiplex_blocks_subscription(
...
+                                panic!("Finalized block stream send gave error {e:?}");
+                            }
+                        } else {
+                            debug!("finalized block meta received for blockhash {} which was never seen or already emitted", blockhash);
+                        }
+                    },
+                    _ = cleanup_tick.tick() => {
+                        let size_before = recent_confirmed_blocks.len();
+                        recent_confirmed_blocks.retain(|_blockhash, block| {
+                            block.slot > current_slot - 100 // must be greater than finalized slot distance (32)
```
godmodegalactus, line 138: Not a good idea, In some cases it takes time to finalize like if there are some complicated bank forks. So current_slot should be last_finalized_slot

godmodegalactus (changes requested):

cluster-endpoints/src/grpc_multiplex.rs
grooviegermanikus: ah! very good point. my logic was very wrong.

grooviegermanikus: should be fixed now

godmodegalactus (dismissed):

godmodegalactus (dismissed):

PR #291, opened 2024-01-19
Making block subscription processed and moving confirmed block subscr…

godmodegalactus:
…iption to meta

### DIFF
--- a/cluster-endpoints/src/grpc_multiplex.rs
+++ b/cluster-endpoints/src/grpc_multiplex.rs
@@ -169,14 +169,15 @@ pub fn create_grpc_multiplex_blocks_subscription(
                             let processed_block = processed_block.expect("processed block from stream");
                             trace!("got processed block {} with blockhash {}",
                                 processed_block.slot, processed_block.blockhash.clone());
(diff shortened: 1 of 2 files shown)


cluster-endpoints/src/grpc_multiplex.rs
```diff
@@ -50,6 +51,67 @@ impl FromYellowstoneExtractor for BlockMetaHashExtractor {
...
+
+    let mut tasks = Vec::new();
+    let mut streams = vec![];
+    for grpc_source in grpc_sources {
+        let (block_sender, block_reciever) = async_channel::unbounded();
+        tasks.push(create_block_processing_task(
+            grpc_source.grpc_addr.clone(),
+            grpc_source.grpc_x_token.clone(),
+            block_sender,
+            yellowstone_grpc_proto::geyser::CommitmentLevel::Confirmed,
```
grooviegermanikus, line 68: shouldn't that be _processed_ or am I missing something?

godmodegalactus: you are right it is not correct

godmodegalactus: Will fix asap

PR #292, opened 2024-01-19
Updating the transport config to match with solana endpoint

godmodegalactus:


services/src/quic_connection_utils.rs
```diff
@@ -79,6 +79,18 @@ impl QuicConnectionUtils {
...
+
+        const DATAGRAM_RECEIVE_BUFFER_SIZE: usize = 1024 * 1024 * 1024;
+        const DATAGRAM_SEND_BUFFER_SIZE: usize = 1024 * 1024 * 1024;
+        const INITIAL_MAXIMUM_TRANSMISSION_UNIT: u16 = MINIMUM_MAXIMUM_TRANSMISSION_UNIT;
+        const MINIMUM_MAXIMUM_TRANSMISSION_UNIT: u16 = 1280;
+        transport_config.datagram_receive_buffer_size(Some(DATAGRAM_RECEIVE_BUFFER_SIZE));
+        transport_config.datagram_send_buffer_size(DATAGRAM_SEND_BUFFER_SIZE);
+        transport_config.initial_mtu(INITIAL_MAXIMUM_TRANSMISSION_UNIT);
+        transport_config.max_concurrent_bidi_streams(VarInt::from(0u8));
+        transport_config.max_concurrent_uni_streams(VarInt::from(0u8));
```
grooviegermanikus, line 91: are you sure that does work for us?

services/src/quic_connection_utils.rs
```diff
@@ -79,6 +79,18 @@ impl QuicConnectionUtils {
         let timeout = IdleTimeout::try_from(Duration::from_secs(1)).unwrap();
         transport_config.max_idle_timeout(Some(timeout));
         transport_config.keep_alive_interval(Some(Duration::from_millis(500)));
+
+        const DATAGRAM_RECEIVE_BUFFER_SIZE: usize = 1024 * 1024 * 1024;
```
grooviegermanikus, line 83: that's one 1GB !

services/src/quic_connection_utils.rs
```diff
@@ -79,6 +79,18 @@ impl QuicConnectionUtils {
...
+        const DATAGRAM_SEND_BUFFER_SIZE: usize = 1024 * 1024 * 1024;
+        const INITIAL_MAXIMUM_TRANSMISSION_UNIT: u16 = MINIMUM_MAXIMUM_TRANSMISSION_UNIT;
+        const MINIMUM_MAXIMUM_TRANSMISSION_UNIT: u16 = 1280;
+        transport_config.datagram_receive_buffer_size(Some(DATAGRAM_RECEIVE_BUFFER_SIZE));
+        transport_config.datagram_send_buffer_size(DATAGRAM_SEND_BUFFER_SIZE);
+        transport_config.initial_mtu(INITIAL_MAXIMUM_TRANSMISSION_UNIT);
```
grooviegermanikus, line 89: not sure if setting the initial value will prevent that heuristic will move the packet size up.
**would expect to have a max (or flex) size**

grooviegermanikus (changes requested):

services/src/quic_connection_utils.rs
grooviegermanikus: this?
image

services/src/quic_connection_utils.rs
godmodegalactus: should work

services/src/quic_connection_utils.rs
godmodegalactus: reduced to 64 MB

grooviegermanikus (approved):
lgtm

PR #305, opened 2024-01-26
reduce log for grpc_multiplex: finalized block meta received for blockhash 

grooviegermanikus:


godmodegalactus:
It is an error and should not be treated as warning

grooviegermanikus:
k. so we need to fix. will close the PR

PR #310, opened 2024-01-31
Calculating prio fees by account

godmodegalactus:


lite-rpc/src/bridge.rs
```diff
@@ -597,4 +615,41 @@ impl LiteRpcServer for LiteBridge {
 
         Ok(())
     }
+
+    async fn get_latest_account_priofees(
+        &self,
+        account: String,
+        method: Option<PrioritizationFeeCalculationMethod>,
```
grooviegermanikus, line 622: would not use an enum+optional. couldn't you just define enum with appropriate variants?

lite-rpc/src/bridge.rs
```diff
@@ -597,4 +615,41 @@ impl LiteRpcServer for LiteBridge {
...
+            ))
+        }
+    }
+
+    async fn latest_account_priofees_subscribe(
+        &self,
+        _pending: PendingSubscriptionSink,
+        _account: String,
+    ) -> SubscriptionResult {
+        todo!()
```
grooviegermanikus, line 653: todo?

lite-rpc/src/rpc.rs
```diff
@@ -237,9 +238,22 @@ pub trait LiteRpc {
...
+    async fn get_latest_block_priofees(
+        &self,
+        method: Option<PrioritizationFeeCalculationMethod>,
+    ) -> crate::rpc::Result<RpcResponse<PrioFeesStats>>;
 
     /// subscribe to prio fees distribution per block; uses confirmation level "confirmed"
     #[subscription(name = "blockPrioritizationFeesSubscribe" => "blockPrioritizationFeesNotification", unsubscribe="blockPrioritizationFeesUnsubscribe", item=PrioFeesStats)]
     async fn latest_block_priofees_subscribe(&self) -> SubscriptionResult;
+
+    #[method(name = "getLatestAccountsPrioFees")]
```
grooviegermanikus, line 250: nit: shouldn't it be _Account_ and not _Accounts_ ?

prioritization_fees/src/block_priofees.rs
```diff
@@ -29,13 +26,33 @@ pub struct PrioFeesService {
...
+        lock.last_key_value()
+            .map(|(slot, value)| (*slot, value.calculate_stats()))
+    }
+
+    pub async fn get_last_n_priofees_aggregate(&self, nb: usize) -> Option<(Slot, PrioFeesStats)> {
+        let lock = self.block_fees_store.recent.read().await;
+        let last_slot = match lock.last_key_value().map(|x| *x.0) {
+            Some(slot) => slot,
+            None => {
+                return None;
```
grooviegermanikus, line 38: maybe you should also abort if there are less than `nb` items in the map?

prioritization_fees/src/prioritization_fee_calculation_method.rs
```diff
@@ -0,0 +1,9 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "camelCase")]
+pub enum PrioritizationFeeCalculationMethod {
+    #[default]
+    Latest,
+    Last(usize),
```
grooviegermanikus, line 8: maybe we can find a more clear naming .. Latest..Last is not really catchy

prioritization_fees/src/prioritization_fee_data.rs
```diff
@@ -0,0 +1,63 @@
...
+        Self {
+            priority: value.0,
+            cu_consumed: value.1,
+        }
+    }
+}
+
+#[derive(Default, Clone)]
+pub struct BlockPrioData {
+    pub tx_prioritization: Vec<PrioFeesData>,
```
grooviegermanikus, line 23: nit: why do you use the term _prioritization_? the list contains the raw prio fees which have impact on the prioritization but that's not the same

prioritization_fees/src/account_prio_service.rs
```diff
@@ -0,0 +1,61 @@
+use solana_lite_rpc_core::types::BlockStream;
+use solana_sdk::pubkey::Pubkey;
+use tokio::task::JoinHandle;
+
+use crate::{account_priofees::AccountPrioStore, rpc_data::AccountPrioFeesStats};
+use tokio::sync::broadcast::error::RecvError::{Closed, Lagged};
+
+#[derive(Clone)]
```
grooviegermanikus, line 8: cloning service + store feels bit wrong but maybe that's just my preference

prioritization_fees/src/account_priofees.rs
```diff
@@ -0,0 +1,220 @@
...
+                match accounts_by_prioritization_write.get_mut(write_lock) {
+                    Some(acc_vec) => {
+                        acc_vec.push(value);
+                    }
+                    None => {
+                        accounts_by_prioritization_write.insert(*write_lock, vec![value]);
+                    }
+                }
+
+                match accounts_by_prioritization_read_write.get_mut(write_lock) {
```
grooviegermanikus, line 66: c&p: should be _read_lock_

prioritization_fees/src/account_priofees.rs
```diff
@@ -0,0 +1,220 @@
...
+                    None => {
+                        accounts_by_prioritization_read_write.insert(*write_lock, vec![value]);
+                    }
+                }
+            }
+
+            for readlock in &transaction.readable_accounts {
+                match accounts_by_prioritization_read_write.get_mut(readlock) {
+                    Some(acc_vec) => {
+                        acc_vec.push(value);
```
grooviegermanikus, line 79: there will be two items added to vec if `for write_lock` and `for readlock` is matching. is that correct?

prioritization_fees/src/account_priofees.rs
```diff
@@ -0,0 +1,220 @@
...
+                    };
+                    prio_fee.stats_by_slot.insert(slot, block_prio_data);
+                    self.account_by_prio_fees_writeonly
+                        .insert(account, prio_fee);
+                }
+            }
+        }
+
+        // cleanup old data
+        let last_slot = produced_block
```
grooviegermanikus, line 141: `last_slot` is very misleading (even worse: there is `self.last_slot` with completely different semantics from what I understand).

maybe call this one `min_slot_to_retain`

prioritization_fees/src/account_priofees.rs
```diff
@@ -0,0 +1,220 @@
...
+            .for_each(cleanup_functor);
+        self.last_slot
+            .store(slot, std::sync::atomic::Ordering::Relaxed);
+
+        self.account_by_prio_fees_writeonly
+            .iter_mut()
+            .for_each(cleanup_functor);
+    }
+
+    pub fn get_latest_stats(&self, account: &Pubkey) -> (u64, AccountPrioFeesStats) {
```
grooviegermanikus, line 163: u64 -> Slot?

grooviegermanikus (changes requested):
some minor comments

prioritization_fees/src/account_prio_service.rs
godmodegalactus: Should be ok, the store is not cloned just a pointer.

prioritization_fees/src/account_priofees.rs
godmodegalactus: Yeah one is only write_lock. other is both write_lock and read lock.

prioritization_fees/src/block_priofees.rs
godmodegalactus: will do it later

PR #316, opened 2024-02-02
Implementing missing endpoints

godmodegalactus:


lite-rpc/src/bridge.rs
```diff
@@ -323,9 +352,34 @@ impl LiteRpcServer for LiteBridge {
...
+        // This method will get the latest global and account prioritization fee stats and then send the maximum p75
+        let accounts = pubkey_strs
+            .iter()
+            .filter_map(|pubkey| Pubkey::from_str(pubkey).ok())
+            .collect_vec();
+
+        let global_prio_fees = self.prio_fees_service.get_latest_priofees().await;
+        let max_p75 = global_prio_fees
+            .map(|(_, fees)| {
+                let fees = fees.get_percentile(0.75).unwrap_or_default();
```
aniketfuryrocks, line 366: Could make a constant 0.75

lite-rpc/src/bridge.rs
```diff
@@ -323,9 +352,34 @@ impl LiteRpcServer for LiteBridge {
...
+                std::cmp::max(fees.0, fees.1)
+            })
+            .unwrap_or_default();
+
+        let mut ret: Vec<RpcPrioritizationFee> = vec![];
+        for account in accounts {
+            let (slot, stats) = self.account_priofees_service.get_latest_stats(&account);
+            let stat = stats.all_stats.get_percentile(0.75).unwrap_or_default();
```
aniketfuryrocks, line 374: Could make a constant 0.75

lite-rpc/src/bridge.rs
```diff
@@ -323,9 +352,34 @@ impl LiteRpcServer for LiteBridge {
...
+        let max_p75 = global_prio_fees
+            .map(|(_, fees)| {
+                let fees = fees.get_percentile(0.75).unwrap_or_default();
+                std::cmp::max(fees.0, fees.1)
+            })
+            .unwrap_or_default();
+
+        let mut ret: Vec<RpcPrioritizationFee> = vec![];
+        for account in accounts {
+            let (slot, stats) = self.account_priofees_service.get_latest_stats(&account);
```
aniketfuryrocks, line 373: I see this function uses `unwrap_or_default`. How does it work if the account does not exist ?

lite-rpc/src/bridge.rs
```diff
@@ -323,9 +352,34 @@ impl LiteRpcServer for LiteBridge {
...
+        let global_prio_fees = self.prio_fees_service.get_latest_priofees().await;
+        let max_p75 = global_prio_fees
+            .map(|(_, fees)| {
+                let fees = fees.get_percentile(0.75).unwrap_or_default();
+                std::cmp::max(fees.0, fees.1)
+            })
+            .unwrap_or_default();
+
+        let mut ret: Vec<RpcPrioritizationFee> = vec![];
+        for account in accounts {
```
aniketfuryrocks, line 372: Could just use accounts.map here. Prevents creating a Vec. Otherwise specify capacity by accounts length to ret

lite-rpc/src/bridge.rs
```diff
@@ -440,8 +494,42 @@ impl LiteRpcServer for LiteBridge {
         todo!()
     }
 
-    async fn slot_subscribe(&self, _pending: PendingSubscriptionSink) -> SubscriptionResult {
-        todo!()
+    async fn slot_subscribe(&self, pending: PendingSubscriptionSink) -> SubscriptionResult {
+        let sink = pending.accept().await?;
+        let mut slot_stream = self.slot_stream.resubscribe();
+        tokio::spawn(async move {
```
aniketfuryrocks, line 500: Doesn't look good. Any one will break down lite-rpc with a lot of empty requests. Preferred moving this to a `Semaphore` implementation using channels. 

1. Maintain one `slot_stream`.
2. Poll if there is a slot_subscribe request.
3. Respond to all requests in a single loop.
4. Sleep for a while if no requests come in or wait on the channel.

Prevents spawning multiple tokio::spawn

aniketfuryrocks (changes requested):

core/src/stores/block_information_store.rs
```diff
@@ -164,4 +166,14 @@ impl BlockInformationStore {
             None => (false, latest_block.slot),
         }
     }
+
+    pub fn get_block_info_by_slot(&self, slot: u64) -> Option<BlockInformation> {
```
grooviegermanikus, line 170: `blocks: Arc<DashMap<String, BlockInformation>>`
^^please provide a comment what "String" is (or better, use a custom type)

core/src/stores/block_information_store.rs
```diff
@@ -164,4 +166,14 @@ impl BlockInformationStore {
...
+        self.blocks.iter().find_map(|iter| {
+            if iter.slot == slot {
```
grooviegermanikus, line 172: pls provide a comment, why that approach is fast enough like "fast enough because we are iterating only a small map of around 100 entries"

lite-rpc/src/bridge.rs
```diff
@@ -178,7 +183,13 @@ impl LiteRpcServer for LiteBridge {
     }
 
     async fn get_cluster_nodes(&self) -> crate::rpc::Result<Vec<RpcContactInfo>> {
-        todo!()
+        Ok(self
+            .data_cache
+            .cluster_info
+            .cluster_nodes
+            .iter()
+            .map(|v| v.value().as_ref().clone())
```
grooviegermanikus, line 191: do not use RpcContactInfo type in store - the type is RPC- specific DTO

lite-rpc/src/bridge.rs
```diff
@@ -194,12 +205,30 @@ impl LiteRpcServer for LiteBridge {
         Ok(slot)
     }
 
-    async fn get_block_height(&self, _config: Option<RpcContextConfig>) -> crate::rpc::Result<u64> {
-        todo!()
+    async fn get_block_height(&self, config: Option<RpcContextConfig>) -> crate::rpc::Result<u64> {
+        let commitment_config = config
+            .map(|x| x.commitment)
+            .unwrap_or_default()
```
grooviegermanikus, line 211: use flat_map instead of double-unwrap

lite-rpc/src/bridge.rs
```diff
@@ -323,9 +352,34 @@ impl LiteRpcServer for LiteBridge {
...
     async fn get_recent_prioritization_fees(
         &self,
-        _pubkey_strs: Option<Vec<String>>,
+        pubkey_strs: Vec<String>,
     ) -> crate::rpc::Result<Vec<RpcPrioritizationFee>> {
-        todo!()
+        // This method will get the latest global and account prioritization fee stats and then send the maximum p75
+        let accounts = pubkey_strs
+            .iter()
+            .filter_map(|pubkey| Pubkey::from_str(pubkey).ok())
```
grooviegermanikus, line 360: shouldn't we return with an error if the account key is not in correct format?

lite-rpc/src/bridge.rs
godmodegalactus: Tokio is made to have multiple tasks, I guess should fine like this instead of making something more complex.

lite-rpc/src/bridge.rs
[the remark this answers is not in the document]
godmodegalactus: These are account stats.

prioritization_fees/src/rpc_data.rs
```diff
@@ -23,6 +23,29 @@ pub struct PrioFeesStats {
     pub cu_consumed: TxAggregateStats,
 }
 
+impl PrioFeesStats {
+    pub fn get_percentile(&self, percentile: f32) -> Option<(u64, u64)> {
+        let index_tx = self
+            .by_tx_percentiles
+            .iter()
```
grooviegermanikus, line 30: could be simplified with this `a.iter().position(|&x| x == 2)`

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position

lite-rpc/src/bridge.rs
godmodegalactus: Tokio.rs states 
```
Tasks in Tokio are very lightweight. Under the hood, they require only a single allocation and 64 bytes of memory. Applications should feel free to spawn thousands, if not millions of tasks.
```

prioritization_fees/src/rpc_data.rs
```diff
@@ -23,6 +23,29 @@ pub struct PrioFeesStats {
...
+            .enumerate()
+            .find(|(_, v)| **v == percentile)
+            .map(|(ix, _)| ix);
+        let index_cu = self
+            .by_cu_percentiles
+            .iter()
+            .enumerate()
+            .find(|(_, v)| **v == percentile)
+            .map(|(ix, _)| ix);
+        if let Some(index_tx) = index_tx {
```
grooviegermanikus, line 40: `if let (Some(index_tx), Some(index_cu)) = (index_tx, index_cu) ...` should work and looks a bit cleaner

grooviegermanikus (changes requested):
the vector account_keys thing is IMO critical - the rest are small improvements

lite-rpc/src/bridge.rs
godmodegalactus: We already store it in this form. There is no real reason to convert to any other form.

lite-rpc/src/bridge.rs
godmodegalactus: There is no method called flat_map

grooviegermanikus (approved):
LGMT

lite-rpc/src/bridge.rs
aniketfuryrocks: I don't agree. Previous experiences say this will end badly.

PR #319, opened 2024-02-05
Adding support for address lookup tables

godmodegalactus:


prioritization_fees/src/account_priofees.rs
```diff
@@ -23,37 +26,67 @@ pub struct AccountPrioStore {
...
             account_by_prio_fees_writeonly: Arc::new(DashMap::new()),
             number_of_slots_to_save,
             last_slot: Arc::new(AtomicU64::new(0)),
+            address_lookup_tables_impl,
         }
     }
 
-    pub fn update(&self, produced_block: &ProducedBlock) -> AccountPrioFeesUpdateMessage {
+    pub async fn update(&self, produced_block: &ProducedBlock) -> AccountPrioFeesUpdateMessage {
         // sort by ascending order
```
grooviegermanikus, line 47: there's a `.rev()` further down; doesn't that change the sort order to _descending_?

address_lookup_tables/src/address_lookup_table_store.rs
```diff
@@ -0,0 +1,218 @@
...
+            Some(account) => {
+                self.save_account(address, &account.data);
+            }
+            None => {
+                log::error!("Cannot find address lookup table {}", address.to_string());
+            }
+        }
+    }
+
+    async fn load_accounts(&self, alt: &Pubkey, accounts: &[u8]) -> Option<Vec<Pubkey>> {
```
grooviegermanikus, line 137: maybe comment what `accounts` is or rename it

grooviegermanikus, line 137: the returned vec must be same size as the requested _accounts_ indices -right? maybe the return value should be `Option<Vec<(usize, Pubkey)>>`

address_lookup_tables/src/address_lookup_table_store.rs
```diff
@@ -0,0 +1,218 @@
...
+                    .collect_vec(),
+            ),
+            None => {
+                log::error!("address lookup table {} was not found", alt);
+                None
+            }
+        }
+    }
+
+    pub async fn get_accounts(&self, alt: &Pubkey, accounts: &[u8]) -> Vec<Pubkey> {
```
grooviegermanikus, line 153: maybe return type should be `Vec<(usize, Pubkey)>`

address_lookup_tables/src/address_lookup_table_store.rs
```diff
@@ -0,0 +1,218 @@
...
+                vec![]
+            }
+        }
+    }
+
+    pub fn serialize_binary(&self) -> Vec<u8> {
+        bincode::serialize::<BinaryALTData>(&BinaryALTData::new(&self.map)).unwrap()
+    }
+
+    pub fn load_binary(&self, binary_data: Vec<u8>) {
```
grooviegermanikus, line 168: maybe rename to _load_from_binary_.
pls clarify if the method is designed to get called only once or multiple times

address_lookup_tables/src/address_lookup_table_store.rs
```diff
@@ -0,0 +1,218 @@
...
+        let binary_alt_data = bincode::deserialize::<BinaryALTData>(&binary_data).unwrap();
+        for (alt, accounts) in binary_alt_data.data.iter() {
+            self.map.insert(*alt, accounts.clone());
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct BinaryALTData {
+    pub data: Vec<(Pubkey, Vec<Pubkey>)>,
```
grooviegermanikus, line 178: can you avoid _pub_ here? the internal structure should be hidden

address_lookup_tables/src/address_lookup_table_store.rs
```diff
@@ -0,0 +1,218 @@
...
+}
+
+impl BinaryALTData {
+    pub fn new(map: &Arc<DashMap<Pubkey, Vec<Pubkey>>>) -> Self {
```
grooviegermanikus, line 182: nit: would not call that _new_. it's an intermediate mapper function. could be map_from() or so

address_lookup_tables/src/address_lookup_table_store.rs
```diff
@@ -0,0 +1,218 @@
...
+                        Ok(multiple_accounts) => {
+                            for (index, acc) in multiple_accounts.value.iter().enumerate() {
+                                if let Some(acc) = acc {
+                                    this.save_account(&batch[index], &acc.data);
+                                }
+                            }
+                        }
+                        Err(e) => {
+                            log::error!(
+                                "error loading {} alts with error {}",
```
grooviegermanikus, line 56: pls write in log, what you do in case of error i.e. "skipping lookup for batch N and continue"

address_lookup_tables/src/address_lookup_table_store.rs
```diff
@@ -0,0 +1,218 @@
...
+                );
+                None
+            }
+        };
+        match account {
+            Some(account) => {
+                self.save_account(address, &account.data);
+            }
+            None => {
+                log::error!("Cannot find address lookup table {}", address.to_string());
```
grooviegermanikus, line 132: pls log what you do here  - i.e. "skip reload of account foo and continue"
would use _warn_ level

grooviegermanikus (changes requested):
some minor changes requested

address_lookup_tables/src/address_lookup_table_store.rs
godmodegalactus: Nah vec Pubkey is fine.

address_lookup_tables/src/address_lookup_table_store.rs
godmodegalactus: Vec Pubkey is fine no use of usize in anycase.

address_lookup_tables/src/address_lookup_table_store.rs
godmodegalactus: There is no method like that, dashmap iterator is a special kind of iterator.

address_lookup_tables/src/address_lookup_table_store.rs
godmodegalactus: Let it be error for now, to see how often it occurs and then could be changed to warn later when it is too verbose

PR #322, opened 2024-02-09
implement simple binary that dumps the effective config

grooviegermanikus:
Dump the Config effectively used for Lite RPC.
Config became quite cumbersome with ENV+config.json+defaults.

> RUST_LOG=info cargo run --bin lite-rpc-dumpconfig

output is provided on `stdout` and `info logger`


grooviegermanikus:
missing Dockefile

grooviegermanikus:
gm does not want it

PR #331, opened 2024-02-16
validate LITE_RPC_HTTP_ADDR in config

Lou-Kamades:


lite-rpc/src/cli.rs
```diff
@@ -124,6 +125,13 @@ impl Config {
         config.lite_rpc_http_addr =
             env::var("LITE_RPC_HTTP_ADDR").unwrap_or(config.lite_rpc_http_addr);
 
+        assert!(
+            TcpListener::bind(config.lite_rpc_http_addr.clone())
+                .await
+                .is_ok(),
```
grooviegermanikus, line 131: please do not use assert for this.

`TcpListener::bind(config.lite_rpc_http_addr.clone())
                 .await.expect("Server address and port for HTTP server");`

lite-rpc/src/cli.rs
```diff
@@ -124,6 +125,13 @@ impl Config {
...
+            "invalid LITE_RPC_HTTP_ADDR"
+        );
+
         config.lite_rpc_ws_addr = env::var("LITE_RPC_WS_ADDR").unwrap_or(config.lite_rpc_ws_addr);
```
grooviegermanikus, line 135: pls do the same check for WS address

grooviegermanikus (changes requested):
pls use different check

grooviegermanikus (approved):
LGTM

PR #335, opened 2024-02-22
Fixing possible memory issues and adding metrics for account prio fees

godmodegalactus:


prioritization_fees/src/account_priofees.rs
```diff
@@ -16,6 +17,17 @@ use crate::{
     rpc_data::{AccountPrioFeesStats, AccountPrioFeesUpdateMessage},
 };
 
+lazy_static::lazy_static! {
+    static ref WRITE_ONLY_ACCOUNTS_IN_PRIO_CACHE: GenericGauge<prometheus::core::AtomicI64> =
```
grooviegermanikus, line 21: we should call the "thing" _priofees_ and not _prio_,

prioritization_fees/src/account_priofees.rs
```diff
@@ -178,24 +190,41 @@ impl AccountPrioStore {
...
+            .account_by_prio_fees_all
+            .iter()
+            .map(|v| v.stats_by_slot.len())
+            .sum();
+        let dp_writeonly: usize = self
+            .account_by_prio_fees_writeonly
+            .iter()
+            .map(|v| v.stats_by_slot.len())
+            .sum();
+        NUMBER_OF_PRIO_DATA_POINTS.set((dp_all + dp_writeonly) as i64);
```
grooviegermanikus, line 223: is that correct? _all_ should include _writeonly_ - no?

prioritization_fees/src/account_priofees.rs
```diff
@@ -178,24 +190,41 @@ impl AccountPrioStore {
         let min_slot_to_retain = produced_block
             .slot
             .saturating_sub(self.number_of_slots_to_save as u64);
-        let cleanup_functor = |mut iter: RefMutMulti<'_, Pubkey, AccountPrio>| {
+        let cleanup_functor = |iter: &mut AccountPrio| {
             while let Some((k, _)) = iter.stats_by_slot.first_key_value() {
```
grooviegermanikus, line 194: you can use the BtreeMap.iter() which gives the entries sorted by key

grooviegermanikus (changes requested):
smal changes

prioritization_fees/src/account_priofees.rs
godmodegalactus: No we mangage for both all and writeonly

prioritization_fees/src/account_priofees.rs
godmodegalactus: Yes that is why i get the first element compare and pop it out. In C++ i know how to iterate and remove multiple first elements. But in rust i cannot find the way.

PR #338, opened 2024-02-26
Implementing account on demand

godmodegalactus:


lite-rpc/src/bridge.rs
```diff
@@ -563,11 +563,7 @@ impl LiteRpcServer for LiteBridge {
                         ui_accounts.push(ui_account);
                     }
                     Err(_) => {
-                        // internal error while fetching multiple accounts
-                        return Err(jsonrpsee::types::error::ErrorCode::ServerError(
-                            RpcErrors::AccountNotFound as i32,
-                        )
-                        .into());
+                        ui_accounts.push(None);
```
grooviegermanikus, line 566: should assert ui_accounts.len()==pubkey_strs.len();
we should do that for all vector/batched request-response pairs!

lite-rpc/src/bridge.rs
```diff
@@ -563,11 +563,7 @@ impl LiteRpcServer for LiteBridge {
                         ui_accounts.push(ui_account);
```
grooviegermanikus, line 563: `lot > max_slot` << this is business logic and cannot be in bridge.rs

lite-rpc/src/bridge_pubsub.rs
```diff
@@ -263,18 +264,24 @@ impl LiteRpcPubSubServer for LitePubSubBridge {
...
                             lagged
                         );
                         continue 'recv_loop;
                     }
-                    Err(Closed) => {
+                    Ok(Err(Closed)) => {
                         log::error!("failed to receive block, sender closed - aborting");
                         return;
                     }
+                    Err(_) => {
```
grooviegermanikus, line 279:
```suggestion
                    Err(_elapsed) => {
```

lite-rpc/src/bridge_pubsub.rs
```diff
@@ -236,11 +236,12 @@ impl LiteRpcPubSubServer for LitePubSubBridge {
...
-                match account_fees_stream.recv().await {
-                    Ok(AccountPrioFeesUpdateMessage {
+                match tokio::time::timeout(Duration::from_secs(1), account_fees_stream.recv()).await
+                {
+                    Ok(Ok(AccountPrioFeesUpdateMessage {
                         slot,
                         accounts_data,
-                    }) => {
+                    })) => {
                         if let Some(account_data) = accounts_data.get(&account) {
```
grooviegermanikus, line 245: _accounts_data_ is a quite bland variable name

lite-rpc/src/bridge_pubsub.rs
```diff
@@ -351,20 +360,26 @@ impl LiteRpcPubSubServer for LitePubSubBridge {
...
                         continue;
                     }
-                    Err(Closed) => {
+                    Ok(Err(Closed)) => {
                         log::error!(
                             "failed to receive account notifications, sender closed - aborting"
                         );
                         return;
                     }
+                    Err(_) => {
```
grooviegermanikus, line 377:
```suggestion
                    Err(_elapsed) => {
```

lite-rpc/src/main.rs
```diff
@@ -153,15 +156,24 @@ pub async fn start_lite_rpc(args: Config, rpc_client: Arc<RpcClient>) -> anyhow:
         vec![]
     };
 
+    if enable_accounts_on_demand_accounts_service {
+        log::info!("Accounts on demand service is enabled");
```
grooviegermanikus, line 160: nit: should also log if it's disabled

lite-rpc/src/main.rs
```diff
@@ -196,7 +208,20 @@ pub async fn start_lite_rpc(args: Config, rpc_client: Arc<RpcClient>) -> anyhow:
         let inmemory_account_storage: Arc<dyn AccountStorageInterface> =
             Arc::new(InmemoryAccountStore::new());
         const MAX_CONNECTIONS_IN_PARALLEL: usize = 10;
-        let account_service = AccountService::new(inmemory_account_storage);
+        let (account_notification_sender, _) = tokio::sync::broadcast::channel(256);
```
grooviegermanikus, line 211: comment the expected size of a single message in the channel + explain the timing characteristics of producer and consumer

lite-rpc/src/main.rs
```diff
@@ -196,7 +208,20 @@ pub async fn start_lite_rpc(args: Config, rpc_client: Arc<RpcClient>) -> anyhow:
...
+
+        let account_storage = if enable_accounts_on_demand_accounts_service {
+            Arc::new(AccountsOnDemand::new(
```
grooviegermanikus, line 214: is that correct? the AccountsOnDemand is an "add-on" on top if the main storage (in memory here) and uses a delegation pattern to implement a smart adapter with geyser

accounts/src/inmemory_account_store.rs
```diff
@@ -1,21 +1,17 @@
...
 use solana_rpc_client_api::filter::RpcFilterType;
 use solana_sdk::{pubkey::Pubkey, slot_history::Slot};
 use std::collections::BTreeMap;
-use tokio::sync::RwLock;
-
-use crate::account_store_interface::AccountStorageInterface;
 
 #[derive(Clone, Default)]
 pub struct AccountDataByCommitment {
+    pub pk: Pubkey,
```
grooviegermanikus, line 14: pls use consistent abbreviations

accounts/src/inmemory_account_store.rs
```diff
@@ -1,21 +1,17 @@
...
     pub processed_accounts: BTreeMap<Slot, AccountData>,
```
grooviegermanikus, line 15: pls comment how many entries we expect here and explain the background

core/src/structures/account_data.rs
```diff
@@ -7,7 +9,7 @@ use crate::commitment_utils::Commitment;
 #[derive(Clone, Eq, PartialEq, Debug)]
```
grooviegermanikus, line 9: how is equality defined here? by pubkey or pubkey+slot? 

what about the **Arc**?

> Two `Arc`s are equal if their inner values are equal, even if they are stored in different allocation.

accounts/src/inmemory_account_store.rs
```diff
@@ -39,20 +35,28 @@ impl AccountDataByCommitment {
...
+            finalized_account: if commitment == Commitment::Finalized {
                 Some(data)
             } else {
                 None
             },
-            finalized_account: None,
         }
     }
 
     pub fn initialize(data: AccountData) -> Self {
```
grooviegermanikus, line 55: pls clarify `initialize` vs `new` - should be renamed accordingly

accounts/src/inmemory_account_store.rs
```diff
@@ -1,21 +1,17 @@
...
 use std::collections::BTreeMap;
-use tokio::sync::RwLock;
-
-use crate::account_store_interface::AccountStorageInterface;
 
 #[derive(Clone, Default)]
 pub struct AccountDataByCommitment {
+    pub pk: Pubkey,
     pub processed_accounts: BTreeMap<Slot, AccountData>,
     pub confirmed_account: Option<AccountData>,
```
grooviegermanikus, line 16: from what I see, all AccountData exist in the BTreeMap keyed by slot.
you could use `Optional<Slot>` for `confirmed_account` and `finalized_account` instead of AccountData

accounts/src/inmemory_account_store.rs
```diff
@@ -240,12 +244,11 @@ impl InmemoryAccountStore {
 
 #[async_trait]
 impl AccountStorageInterface for InmemoryAccountStore {
-    async fn update_account(&self, account_data: AccountData, commitment: Commitment) {
+    async fn update_account(&self, account_data: AccountData, commitment: Commitment) -> bool {
```
grooviegermanikus, line 247: pls clarify what "return bool" indicates

accounts/src/inmemory_account_store.rs
```diff
@@ -74,12 +78,14 @@ impl AccountDataByCommitment {
             .map(|x| x.updated_slot < data.updated_slot)
```
grooviegermanikus, line 78: the naming `updated_slot` is strange. what does it mean in context of the struct?
should only be called `slot`

accounts/src/inmemory_account_store.rs
```diff
@@ -74,12 +78,14 @@ impl AccountDataByCommitment {
...
             .unwrap_or(true);
 
+        let mut updated = false;
         if self.processed_accounts.get(&data.updated_slot).is_none() {
             // processed not present for the slot
             self.processed_accounts
                 .insert(data.updated_slot, data.clone());
```
grooviegermanikus, line 85: this misses the case in this flow:

GIVEN emptyMap, None, None
WHEN
 `.update(1234, confirmed)` for a slot that was not registered before
THEN_WRONG
 map(1234, None, None)
THEN_CORRECT
 map(1234, 1234, None)

accounts/src/inmemory_account_store.rs
```diff
@@ -165,33 +165,37 @@ impl AccountDataByCommitment {
...
-        } else {
+        } else if commitment == Commitment::Finalized {
             // remove processed slot data
-            while self
-                .processed_accounts
-                .first_key_value()
-                .map(|(s, _)| *s)
-                .unwrap_or(u64::MAX)
-                <= slot
+            while self.processed_accounts.len() > 1
```
grooviegermanikus, line 170: not sure but it could be that there are processed blocks later than `slot` which got slashed. so they should also be removed

accounts/src/inmemory_account_store.rs
```diff
@@ -165,33 +165,37 @@ impl AccountDataByCommitment {
...
             None
         }
     }
 }
 
 pub struct InmemoryAccountStore {
     account_store: Arc<DashMap<Pubkey, AccountDataByCommitment>>,
-    confirmed_slots_map: RwLock<BTreeSet<Slot>>,
+    confirmed_slots_map: DashSet<Slot>,
     owner_map_accounts: Arc<DashMap<Pubkey, HashSet<Pubkey>>>,
```
grooviegermanikus, line 191: nit: could be named `accounts_by_owner`

accounts/src/inmemory_account_store.rs
```diff
@@ -260,14 +263,15 @@ impl AccountStorageInterface for InmemoryAccountStore {
                 if let Some(prev_account) = prev_account {
```
grooviegermanikus, line 263: instead of **Occupied, etc**, this should do it:
`self.owner_map_accounts.entry(owner).or_default().insert(account);`

accounts/src/inmemory_account_store.rs
```diff
@@ -240,12 +244,11 @@ impl InmemoryAccountStore {
```
grooviegermanikus, line 244: `update_owner` really looks strange.

this looks like an assertion: `prev_account_data.pubkey == new_account_data.pubkey`

this should fast-return `prev_account_data.account.owner == new_account_data.account.owner`

the finalized case could call a method `self.set_only_account_owner` which is a variant of `self.add_account_owner`

accounts-on-demand/src/accounts_on_demand.rs
```diff
@@ -0,0 +1,228 @@
...
+
+#[async_trait]
+impl AccountStorageInterface for AccountsOnDemand {
+    async fn update_account(&self, account_data: AccountData, commitment: Commitment) -> bool {
+        self.accounts_storage
+            .update_account(account_data, commitment)
+            .await
+    }
+
+    async fn initilize_account(&self, account_data: AccountData) {
```
grooviegermanikus, line 89: typo

accounts/src/inmemory_account_store.rs
```diff
@@ -165,33 +165,37 @@ impl AccountDataByCommitment {
...
+
+            None
+        } else {
             None
         }
     }
 }
 
 pub struct InmemoryAccountStore {
     account_store: Arc<DashMap<Pubkey, AccountDataByCommitment>>,
```
grooviegermanikus, line 189: the size of this map should be tracked with prometheus metric

cluster-endpoints/src/grpc/gprc_accounts_streaming.rs
```diff
@@ -28,10 +27,7 @@ pub fn start_account_streaming_tasks(
 ) -> AnyhowJoinHandle {
```
grooviegermanikus, line 27: we should really avoid Unbounded Channels.

I use this pattern on the .send side to track blocking channels: [example](https://github.com/grooviegermanikus/rust-debug-collections/blob/0a044224c5c8c7712ed3705d16d3191c6e2e63cc/src/tokio_wrapped/mpsc/channels_wrapped.rs#L7)

accounts-on-demand/src/subscription_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+            let processed_commitment = yellowstone_grpc_proto::geyser::CommitmentLevel::Processed;
+
+            let mut subscribe_accounts: HashMap<String, SubscribeRequestFilterAccounts> =
+                HashMap::new();
+
+            let accounts_filters = accounts_filters.read().await.clone();
+            for (index, accounts_filter) in accounts_filters.iter().enumerate() {
+                if !accounts_filter.accounts.is_empty() {
+                    subscribe_accounts.insert(
+                        format!("accounts_{index:?}"),
```
grooviegermanikus, line 98: I like that you prefix the filter keys but could you use a dedicated prefix for accounts and programs?
`account_0` and `program_account_12121221`

cluster-endpoints/src/grpc/gprc_accounts_streaming.rs
```diff
@@ -130,17 +138,17 @@ pub fn start_account_streaming_tasks(
...
                                         owner: Pubkey::new_from_array(owner),
                                         executable: account_data.executable,
                                         rent_epoch: account_data.rent_epoch,
-                                    },
+                                    }),
                                     updated_slot: account.slot,
                                 },
                                 // TODO update with processed commitment / check above
-                                commitment: Commitment::Confirmed,
+                                commitment: Commitment::Processed,
```
grooviegermanikus, line 151: consolidate with `processed_commitment`

cluster-endpoints/src/grpc/gprc_accounts_streaming.rs
```diff
@@ -104,14 +105,21 @@ pub fn start_account_streaming_tasks(
                 blocks: Default::default(),
                 blocks_meta: Default::default(),
                 entry: Default::default(),
-                commitment: Some(confirmed_commitment.into()),
+                commitment: Some(processed_commitment.into()),
                 accounts_data_slice: Default::default(),
                 ping: None,
```
grooviegermanikus, line 110: guess you should not match for Ping later

accounts-on-demand/src/subscription_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+            log::info!(
+                "Accounts on demand subscribing to {}",
+                grpc_config.grpc_addr
+            );
+            let mut client = yellowstone_grpc_client::GeyserGrpcClient::connect(
+                grpc_config.grpc_addr.clone(),
+                grpc_config.grpc_x_token.clone(),
+                None,
+            )
+            .unwrap();
```
grooviegermanikus, line 182: **caution:** this will panic your thread

accounts-on-demand/src/subscription_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+            let mut account_stream = client.subscribe_once2(subscribe_request).await.unwrap();
```
grooviegermanikus, line 183: this will panic your thread in case of error

accounts-on-demand/src/subscription_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+        let _ = self.restart_sender.send(());
+    }
+}
+
+pub fn start_account_streaming_tasks(
+    grpc_config: GrpcSourceConfig,
+    accounts_filters: Arc<RwLock<AccountFilters>>,
+    mut restart_channel: broadcast::Receiver<()>,
+    account_stream_sx: broadcast::Sender<AccountNotificationMessage>,
+) -> AnyhowJoinHandle {
```
grooviegermanikus, line 86: _AnyhowJoinHandle_  is useless here from what I see

cluster-endpoints/src/grpc/gprc_accounts_streaming.rs
```diff
@@ -168,7 +176,7 @@ pub fn create_grpc_account_streaming(
     grpc_sources: Vec<GrpcSourceConfig>,
     accounts_filters: AccountFilters,
 ) -> (AnyhowJoinHandle, AccountStream) {
-    let (account_sender, accounts_stream) = broadcast::channel::<AccountNotificationMessage>(128);
+    let (account_sender, accounts_stream) = broadcast::channel::<AccountNotificationMessage>(1024);
 
     let jh: AnyhowJoinHandle = tokio::spawn(async move {
         loop {
```
grooviegermanikus, line 182: `start_account_streaming_tasks` is leaking tasks when reconnecting

grooviegermanikus (changes requested):
See comments

accounts-on-demand/src/accounts_on_demand.rs
```diff
@@ -0,0 +1,228 @@
...
+        self.accounts_storage.initilize_account(account_data).await
```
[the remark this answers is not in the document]
godmodegalactus: I will create a new PR with all the metrics.

accounts-on-demand/src/subscription_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+    restart_sender: broadcast::Sender<()>,
+}
+
+impl SubscriptionManger {
+    pub fn new(
+        grpc_sources: Vec<GrpcSourceConfig>,
+        accounts_storage: Arc<dyn AccountStorageInterface>,
+        account_notification_sender: Sender<AccountNotificationMessage>,
+    ) -> Self {
+        let (restart_sender, recver) = broadcast::channel(1);
```
grooviegermanikus, line 38: typo

grooviegermanikus, line 38: would use _watch_ instead -> https://docs.rs/tokio/latest/tokio/sync/watch/index.html

> This channel is useful for watching for changes to a value from multiple points in the code base, for example, changes to configuration values.

accounts-on-demand/src/subscription_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+                        }
+                    },
+                }
+            }
+        }
+        Ok(())
+    })
+}
+
+pub fn create_grpc_account_streaming(
```
grooviegermanikus, line 253: we should call methods who spawn a task consistently: `start_foobar_task()`

accounts-on-demand/src/subscription_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+                    Ok(Err(e)) => match e {
+                        broadcast::error::RecvError::Closed => {
+                            log::error!("Account stream channel closed");
+                            break;
+                        }
+                        broadcast::error::RecvError::Lagged(_) => {
+                            continue;
+                        }
+                    },
+                    Err(_) => {
```
grooviegermanikus, line 63:
```suggestion
                    Err(_elapsed) => {
```

accounts-on-demand/src/subscription_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+                match tokio::time::timeout(Duration::from_secs(60), account_stream.recv()).await {
+                    Ok(Ok(message)) => {
+                        let _ = account_notification_sender.send(message.clone());
+                        accounts_storage
+                            .update_account(message.data, message.commitment)
+                            .await;
+                    }
+                    Ok(Err(e)) => match e {
+                        broadcast::error::RecvError::Closed => {
+                            log::error!("Account stream channel closed");
```
grooviegermanikus, line 56: this will stop the tasks and give up on accounts stream - is that, what you want?

accounts-on-demand/src/subscription_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+                            break;
+                        }
+                        broadcast::error::RecvError::Lagged(_) => {
```
grooviegermanikus, line 59: uh, that should write a warning log, because account updates are effectively lost - no?

accounts-on-demand/src/subscription_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+        Ok(())
+    })
+}
+
+pub fn create_grpc_account_streaming(
+    grpc_sources: Vec<GrpcSourceConfig>,
+    accounts_filters: Arc<RwLock<AccountFilters>>,
+    restart_channel: broadcast::Receiver<()>,
+) -> (AnyhowJoinHandle, AccountStream) {
+    let (account_sender, accounts_stream) = broadcast::channel::<AccountNotificationMessage>(128);
```
grooviegermanikus, line 258: IMO a producer component should write into a mpsc. the broadcast "fanout" should be configured by the high-level application logic (i.e. literpc/main.rs)

my private design guideline is: what should happen on backpressure (i.e. when consumer is slower than producer; who decides about the buffer size) and who controls the lifecycle of the channel (who creates it and who destroys it)

accounts-on-demand/src/subscription_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+                    }
+                    UpdateOneof::Ping(_) => {
+                        log::trace!("GRPC Ping accounts stream");
+                    }
+                    _ => {
+                        log::error!("GRPC accounts steam misconfigured");
+                    }
+                };
+
+                match restart_channel.try_recv() {
```
grooviegermanikus, line 232: **be careful**: IMO you should do` loop{ select!`

accounts-on-demand/src/subscription_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+                        broadcast::error::TryRecvError::Closed => {}
+                        broadcast::error::TryRecvError::Lagged(_) => {
+                            // remove message in the queue
+                            let _ = restart_channel.try_recv();
+                            break;
+                        }
+                    },
+                }
+            }
+        }
```
grooviegermanikus, line 248:
```suggestion
        } // -- END main_loop
```

accounts-on-demand/src/subscription_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+        Ok(())
+    })
+}
+
+pub fn create_grpc_account_streaming(
+    grpc_sources: Vec<GrpcSourceConfig>,
+    accounts_filters: Arc<RwLock<AccountFilters>>,
```
grooviegermanikus, line 255: I would rather send the new filters via the **restart_channel**

accounts-on-demand/src/subscription_manager.rs
grooviegermanikus: if you consider using my auto connect it should be fine [huge autoconnect](https://github.com/blockworks-foundation/geyser-grpc-connector/blob/a6ab686fa4607e95550d29f77f08ea074b90e91f/src/grpc_subscription_autoreconnect_tasks.rs#L143)

accounts-on-demand/src/subscription_manager.rs
```diff
@@ -0,0 +1,281 @@
...
+                                        owner: Pubkey::new_from_array(owner),
+                                        executable: account_data.executable,
+                                        rent_epoch: account_data.rent_epoch,
+                                    }),
+                                    updated_slot: account.slot,
+                                },
+                                // TODO update with processed commitment / check above
+                                commitment: Commitment::Processed,
+                            };
+                            if account_stream_sx.send(notification).is_err() {
```
grooviegermanikus, line 217: why are you doing that? the error happens if broadcast has no receivers. that's not an error and it does not indicate a problem with upstream. or am I missing something?

accounts-on-demand/src/accounts_on_demand.rs
```diff
@@ -0,0 +1,228 @@
...
+            program_filters: Arc::new(RwLock::new(vec![])),
+            subscription_manager: SubscriptionManger::new(
+                grpc_sources,
+                accounts_storage,
+                account_notification_sender,
+            ),
+        }
+    }
+
+    pub async fn reset_subscription(&self) {
```
grooviegermanikus, line 53: _reset_subscription_ is a but misleading. to me it reads like "remove all subscription" but actually what is does is _reconfigure_subscriptions_

accounts-on-demand/src/accounts_on_demand.rs
```diff
@@ -0,0 +1,228 @@
...
+            };
+            filters.push(account_filter);
+        }
+
+        self.subscription_manager
+            .update_subscriptions(filters)
+            .await;
+    }
+
+    pub async fn get_filters(&self) -> AccountFilters {
```
grooviegermanikus, line 76: maybe this should not be public

accounts-on-demand/src/accounts_on_demand.rs
```diff
@@ -0,0 +1,228 @@
...
+        match self
+            .accounts_storage
+            .get_account(account_pk, commitment)
+            .await
+        {
+            Some(account_data) => Some(account_data),
+            None => {
+                // account does not exist in account store
+                // first check if we have already subscribed to the required account
+                if !self.accounts_subscribed.contains(&account_pk) {
```
grooviegermanikus, line 103: IMO this is a race condition.
the resubscription happens asynchronously - i.e. the values in `accounts_subscribed` are "subject to subscription" but might not yet arrived in the store. as you skip the RPC lookup in this case the account appears as if it does not exist

accounts-on-demand/src/subscription_manager.rs
grooviegermanikus: I would do microbatching (i.e. wait 500ms or so) and then consume and merge all requested accounts instead of trigger re-subscription per new filter list

grooviegermanikus:
_ui_account_ is quite expensive - should be calculated only `if minimum_context_slot <= account_data.updated_slot`
```
            let ui_account =
                Self::convert_account_data_to_ui_account(&account_data, config.clone());

            // if minimum context slot is not satisfied return Null
            let minimum_context_slot = config
                .as_ref()
                .map(|c| c.min_context_slot.unwrap_or_default())
                .unwrap_or_default();
            if minimum_context_slot <= account_data.updated_slot {
                Ok((account_data.updated_slot, Some(ui_account)))
            } else {
                Ok((account_data.updated_slot, None))
            }
```

grooviegermanikus:
IMO you should also track usage of _accounts_filters_:
* when a subscribed account was last used
* how often it was used

consider implementing unsubscribe for subscribed accounts which where not requested for n minutes

lite-rpc/src/main.rs
godmodegalactus: Yes

core/src/structures/account_data.rs
godmodegalactus: Yes you are right

accounts/src/inmemory_account_store.rs
godmodegalactus: Simpler to copy, anycase underlying data is stored in Arc so account data will not be really copied.

accounts/src/inmemory_account_store.rs
godmodegalactus: When was the account last updated.

accounts/src/inmemory_account_store.rs
godmodegalactus: No because processed accounts should atleast contain the last updated account state.

accounts/src/inmemory_account_store.rs
godmodegalactus: We do different things if account space has been occupied or not.

accounts-on-demand/src/accounts_on_demand.rs
godmodegalactus: It is defined in the trait which aims to warmup the cache.

cluster-endpoints/src/grpc/gprc_accounts_streaming.rs
godmodegalactus: Should be fine for now.

accounts-on-demand/src/subscription_manager.rs
godmodegalactus: On multiple updates would always want last updated

godmodegalactus:
Will add the metrics in next PR.

accounts-on-demand/src/subscription_manager.rs
godmodegalactus: Nah if there are no subscption it means that there are no subscriptions for accounts for the moment. Should not do anything in that case.

godmodegalactus: > Nah if there are no subscption it means that means no one is listening to the geyser stream.

accounts-on-demand/src/accounts_on_demand.rs
godmodegalactus: account_subscribed does not contain the value we add it and reset subscription. This is done because there might be accounts that do not exists so we want to check if these accounts were already subscribed if not then we start get it from RPC and restart the subscription.

PR #344, opened 2024-02-29
account_subscription_into_multiple_channels_and_merge_them

godmodegalactus:


lite-rpc/src/main.rs
```diff
@@ -390,6 +390,8 @@ pub async fn start_lite_rpc(args: Config, rpc_client: Arc<RpcClient>) -> anyhow:
     ));
     drop(slot_notifier);
 
+    println!("LiteRPC started");
```
grooviegermanikus, line 393: pls remove

PR #356, opened 2024-03-11
After fanatids review

godmodegalactus:


services/src/tx_sender.rs
```diff
@@ -25,21 +23,12 @@ lazy_static::lazy_static! {
...
-        "literpc_txs_send_timer",
-        "Time to send transaction batch",
-    ))
-    .unwrap();
     static ref TX_TIMED_OUT: GenericGauge<prometheus::core::AtomicI64> = register_int_gauge!(opts!("literpc_tx_timeout", "Number of transactions that timeout")).unwrap();
     pub static ref TXS_IN_CHANNEL: GenericGauge<prometheus::core::AtomicI64> = register_int_gauge!(opts!("literpc_txs_in_channel", "Transactions in channel")).unwrap();
 
 }
 
-// making 250 as sleep time will effectively make lite rpc send
```
godmodegalactus, line 39: This logic is no longer needed with new tpu client.

services/src/tx_sender.rs
```diff
@@ -133,54 +78,36 @@ impl TxSender {
...
-                                log::error!("Channel Disconnected");
-                                bail!("Channel Disconnected");
-                            }
-                        },
-                        Err(_) => {
-                            break;
-                        }
+            let mut instant = Instant::now();
+            let mut notifications = vec![];
+            while let Some(transaction_info) = recv.recv().await {
```
fanatid, line 83: we can lock the process for an unknown time, better apply timeout + try_recv
something like this
```rust
loop {
    let deadline = Instant::now() + INTERVAL_PER_BATCH_IN_MS;
    let mut notifications = vec![];
    loop {
        let maybe_transaction_info = if deadline.elapsed().is_zer() {
            match tokio::time::timeout_at(deadline, recv.recv()).await {
                Ok(Some(tx_info)) => Some(tx_info),
                Ok(None) => bail!("Tx sender service stopped"),
                Err(_error) => None,
            }
        } else {
            match recv.try_recv() {
                tokio::sync::mpsc::error::TryRecvError::Empty => None,
                tokio::sync::mpsc::error::TryRecvError::Disconnected => bail!("Tx sender service stopped")
            }
        };

        if let Some(transaction_info) = maybe_transaction_info {
            self.forward_txs(&transaction_info).await;

            // send transaction notifications in batches
            if let Some(notifier) = &notifier {
                let forwarded_slot = self.data_cache.slot_cache.get_current_slot();
                let forwarded_local_time = Utc::now();
                let tx_notification = TransactionNotification {
                    signature: transaction_info.signature,
                    recent_slot: transaction_info.slot,
                    forwarded_slot,
                    forwarded_local_time,
                    processed_slot: None,
                    cu_consumed: None,
                    cu_requested: None,
                    quic_response: 0,
                };
                notifications.push(tx_notification);
            }
        } else {
            // send transaction notifications in batches
            if let Some(notifier) = &notifier {
                // send notification for sent transactions
                let _ = notifier.send(NotificationMsg::TxNotificationMsg(notifications));
            }
            break;
        }
    }
}
```

core/src/structures/rotating_queue.rs
```diff
@@ -1,36 +1,30 @@
...
-        item
     }
 
     pub fn get(&self) -> Option<T> {
         if !self.elements.is_empty() {
             let current = self.current.fetch_add(1, Ordering::Relaxed);
-            let index = current % (self.elements.len() as u64);
-            Some(self.elements[index as usize].clone())
+            let index = current % (self.elements.len());
+            Some(self.elements[index].clone())
```
fanatid, line 27:
```suggestion
            self.elements.get(index).cloned()
```

services/src/tx_sender.rs
godmodegalactus: Have addressed your concerns using tokio select, it seems more readable for me.

services/src/tx_sender.rs
```diff
@@ -78,32 +78,51 @@ impl TxSender {
...
-                    notifications.push(tx_notification);
-                    if instant.elapsed() > INTERVAL_PER_BATCH_IN_MS {
-                        // send notification for sent transactions
-                        let _ = notifier.send(NotificationMsg::TxNotificationMsg(
-                            notifications.drain(..).collect_vec(),
-                        ));
-                        instant = Instant::now();
+                    // send notification for sent transactions
+                    let _ = notifier.send(NotificationMsg::TxNotificationMsg(
+                        notifications.drain(..).collect_vec(),
```
fanatid, line 91: `std::mem::take` does not need iterator, but maybe resulted generated code would be same

fanatid (approved):
lgtm

PR #357, opened 2024-03-13
Use CLI instead of binaries for configuring benchmarks

Lou-Kamades:


bench/src/benches/confirmation_rate.rs
```diff
@@ -31,29 +30,25 @@ pub struct RpcStat {
 }
 
 /// TC2 send multiple runs of num_txns, measure the confirmation rate
-#[tokio::main]
-async fn main() -> anyhow::Result<()> {
-    tracing_subscriber::fmt::init();
-
+pub async fn confirmation_rate(
+    payer_path: String,
```
grooviegermanikus, line 34: nit: I'd prefer strong typing where possible. i.e. Path here

PathBuf vs Path: Path only holds a reference to the path string data but doesn't own this data, while PathBuf owns the string data itself

bench/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+        api_load::api_load, confirmation_rate::confirmation_rate,
+        confirmation_slot::confirmation_slot,
+    },
+    tx_size::TxSize,
+};
+use clap::{Parser, Subcommand};
+
+#[derive(Parser, Debug)]
+#[clap(version, about)]
+
```
grooviegermanikus, line 12: do you want to keep sub commands? ffti

grooviegermanikus (commented):
lgtm

grooviegermanikus (commented):
lgtm

grooviegermanikus (approved):
lgtm

PR #361, opened 2024-03-18
add histogram utils

grooviegermanikus:
* histogram logic we used for prio fees
* histogram from another project I found based on buckets


util/src/histogram_nbuckets.rs
```diff
@@ -0,0 +1,23 @@
+pub fn histogram(values: &[f64], bins: usize) -> Vec<(f64, usize)> {
+    assert!(bins >= 2);
+    let mut bucket: Vec<usize> = vec![0; bins];
+
+    let mut min = std::f64::MAX;
+    let mut max = std::f64::MIN;
```
godmodegalactus, line 6: Can use iter().min() and iter().max() instead.

godmodegalactus (changes requested):

PR #362, opened 2024-03-18
restore old benchmarks

Lou-Kamades:


grooviegermanikus:
@Lou-Kamades  FYI renamed the second binary from cli to `benchnew`

PR #367, opened 2024-03-22
use geyser blockmeta to progress finalizedslot

grooviegermanikus:
changes dependency of block_information_store to block stream->block info stream; this should propagate finalized block information faster

the important change is inside the `.listen` method: look for _block_information_store_block_info_ as this is updated from blockinfo_stream

### DIFF
--- a/cluster-endpoints/src/grpc_multiplex.rs
+++ b/cluster-endpoints/src/grpc_multiplex.rs
@@ -9,6 +9,7 @@ use solana_sdk::clock::Slot;
 use solana_sdk::commitment_config::CommitmentConfig;
 
 use solana_lite_rpc_core::solana_utils::hash_from_str;
+use solana_lite_rpc_core::structures::block_info::BlockInfo;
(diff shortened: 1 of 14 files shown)


cluster-endpoints/src/grpc_multiplex.rs
```diff
@@ -211,9 +225,9 @@ pub fn create_grpc_multiplex_blocks_subscription(
             let (processed_block_sender, mut processed_block_reciever) =
                 tokio::sync::mpsc::channel::<ProducedBlock>(10); // experiemental
             let (block_meta_sender_confirmed, mut block_meta_reciever_confirmed) =
-                tokio::sync::mpsc::channel::<BlockMeta>(500);
+                tokio::sync::mpsc::channel::<BlockInfo>(500);
             let (block_meta_sender_finalized, mut block_meta_reciever_finalized) =
```
ckamm, line 229: also rename to block_info_sender/receiver?

ckamm (approved):

PR #382, opened 2024-04-02
optimize mapping of produced block

grooviegermanikus:
will call full block mapper as late as possible

### DIFF
--- a/cluster-endpoints/src/grpc_multiplex.rs
+++ b/cluster-endpoints/src/grpc_multiplex.rs
@@ -49,7 +49,7 @@ fn create_grpc_multiplex_processed_block_task(
     let jh_merging_streams = tokio::task::spawn(async move {
         let mut slots_processed = BTreeSet::<u64>::new();
         let mut last_tick = Instant::now();
-        loop {
+        'recv_loop: loop {
             // recv loop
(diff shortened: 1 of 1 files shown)


cluster-endpoints/src/grpc_multiplex.rs
```diff
@@ -70,16 +70,23 @@ fn create_grpc_multiplex_processed_block_task(
             };
             match blocks_rx_result {
                 Some(Message::GeyserSubscribeUpdate(subscribe_update)) => {
-                    let mapfilter =
-                        map_block_from_yellowstone_update(*subscribe_update, COMMITMENT_CONFIG);
-                    if let Some((slot, produced_block)) = mapfilter {
-                        assert_eq!(COMMITMENT_CONFIG, produced_block.commitment_config);
+                    // note: avoid mapping of full block as long as possible
+                    let extracted_slot = extract_slot_from_yellowstone_update(&subscribe_update);
+                    if let Some(slot) = extracted_slot {
```
ckamm, line 75: minor: avoid one level of indentation with
```
let Some(slot) = extracted-slot else {
    continue 'recv_loop;
}
```

ckamm (approved):

PR #386, opened 2024-04-03
experiment: reduce number of source endpoints to 1

ckamm:
Prediction is that this has no negative effect but reduces memory use a lot.

### DIFF
--- a/services/src/tpu_utils/tpu_connection_manager.rs
+++ b/services/src/tpu_utils/tpu_connection_manager.rs
@@ -249,9 +249,9 @@ impl TpuConnectionManager {
     pub async fn new(
         certificate: rustls::Certificate,
         key: rustls::PrivateKey,
-        fanout: usize,
+        _fanout: usize,
     ) -> Self {
-        let number_of_clients = fanout * 4;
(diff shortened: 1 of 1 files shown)


services/src/tpu_utils/tpu_connection_manager.rs
```diff
@@ -249,9 +249,9 @@ impl TpuConnectionManager {
     pub async fn new(
         certificate: rustls::Certificate,
         key: rustls::PrivateKey,
-        fanout: usize,
+        _fanout: usize,
     ) -> Self {
-        let number_of_clients = fanout * 4;
+        let number_of_clients = 1; // fanout * 4;
```
godmodegalactus, line 254: Do we make it as an argument later then ?

godmodegalactus (approved):

services/src/tpu_utils/tpu_connection_manager.rs
ckamm: yes, I want to see if it works first - and if it does, we can drop the RotatingQueue entirely

PR #399, opened 2024-06-17
moving unable to connection to a node using quic from warning to debug

godmodegalactus:
These days many of nodes have custom scrpts to block traffic which is not originating from gossip node.


grooviegermanikus (approved):
lgtm

